PR #34, opened 2021-08-18
Add clinical trials source

bgyori:
This PR adds a source processor for clinicaltrials.gov. The extraction could be done in several ways, in this approach we model the trials themselves as nodes, and add relations for drugs being tested in a given trial, as well as relations for diseases that a given trial is conducted for.


bgyori:
This is good as an initial implementation but could benefit from a more sophisticated implementation of getting groundings for drugs and diseases by combining the (incomplete) MeSH IDs provided in the data and the ones obtained from grounding. This could be done in a follow-up PR later.

PR #35, opened 2021-08-25
Add more metadata to INDRA edges

bgyori:
This PR adds the evidence count to INDRA edges. @cthoyt I wonder if we should make a bigger change here. Namely, we might want to change the approach of using `stmt_type` as the Relation's type corresponding to INDRA Statement. The reason is that for various path/pattern queries, we want to be able to traverse _any_ INDRA Statement edge, and currently we'd have to enumerate all statement types explicitly to be able to do that.


bgyori:
After some further thought I went ahead with the relation type change to `indra_rel` which is intentionally generic. I added `stmt_type`, `belief` and `evidence_count` as new data entries.

PR #38, opened 2021-09-15
Interpretation

STIRLIN6:
Add client to the Odinson module

### DIFF
--- a/src/indra_cogex/sources/odinson/document.py
+++ b/src/indra_cogex/sources/odinson/document.py
@@ -2,8 +2,12 @@
 interpretations and for processing into this representation as well
 as generating visualizations."""
 
+import glob
 import gzip
 import json
+import tqdm
+import os.path
+
 import networkx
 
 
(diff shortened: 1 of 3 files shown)


src/indra_cogex/sources/odinson/document.py
```diff
@@ -0,0 +1,152 @@
...
+import glob
+import gzip
+import json
+import tqdm
+import os.path
+
+import networkx
+
+
+class Token:
```
cthoyt, line 14: would be a good place to use pydantic

src/indra_cogex/sources/odinson/grammars.py
```diff
@@ -0,0 +1,20 @@
+class Rule:
+    def __init__(self, name, label, type, pattern):
+        self.name = name
+        self.label = label
+        self.type = type
+        self.pattern = pattern
+
+    def compile(self):
```
cthoyt, line 8: why not just call this `__str__`?

src/indra_cogex/sources/odinson/grammars.py
```diff
@@ -0,0 +1,20 @@
+class Rule:
+    def __init__(self, name, label, type, pattern):
+        self.name = name
+        self.label = label
+        self.type = type
+        self.pattern = pattern
```
cthoyt, line 6: you should use [dataclasses](https://docs.python.org/3/library/dataclasses.html) for this, which uses type annotations to automatically write the `__init__` function for you

```suggestion
from dataclasses import dataclass
class Rule:
    name: str
    label: str
    type: str
    pattern: str
```

PR #40, opened 2021-10-13
Add processors for drug indications and side effects

### LINKED ISSUE #39: ChEMBL Indication Processor
This SQL statement can be used with the `chembl_downloader` to get links between chemicals and their indications (either as mesh or EFO/HP) and their maximum clinical phase:

```sql
SELECT
    MOLECULE_DICTIONARY.chembl_id,
    MOLECULE_DICTIONARY.pref_name,
    DRUG_INDICATION.mesh_id,
    DRUG_INDICATION.mesh_heading,
    DRUG_INDICATION.efo_id,
    DRUG_INDICATION.efo_term,
    DRUG_INDICATION.max_phase_for_ind
FROM MOLECULE_DICTIONARY
JOIN DRUG_INDICATION ON MOLECULE_DICTIONARY.molregno == DRUG_INDICATION.molregno
```

There's another column in `MOLECULE_DICTIONARY` called `chebi_par_id` but a spot check of the first 10 rows showed this doesn't have good coverage

cthoyt:
Closes #39

This PR adds a processor over the ChEMBL database based on SQL queries executed by the [`chembl_downloader`](https://github.com/cthoyt/chembl_downloader). It standardizes the ChEMBL identifiers for chemicals and the MeSH identifiers for indications using the INDRA BioOntology. The code that does this is probably generally reusable so I gave it a its own utility function.

This PR also adds the processor for SIDER side effects since it shares some of the standardization code from the ChEMBL processor

### DIFF
--- a/src/indra_cogex/sources/sider/__init__.py
+++ b/src/indra_cogex/sources/sider/__init__.py
@@ -15,9 +15,11 @@ from biomappings import load_mappings
 from tabulate import tabulate
 from tqdm import tqdm
 
+from indra.databases import biolookup_client
 from indra.databases.identifiers import get_ns_id_from_identifiers
 from indra_cogex.representation import Node, Relation, standardize
 from indra_cogex.sources import Processor
+
 
 VERSION = "4.1"
 SUBMODULE = pystow.module("indra", "cogex", "sider", VERSION)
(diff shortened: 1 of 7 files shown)


src/indra_cogex/sources/sider/__init__.py
```diff
@@ -0,0 +1,257 @@
...
+                    self.xrefs[source_prefix][source_id] = target_id
+                else:
+                    self.xrefs[source_prefix] = {
+                        source_id: target_id,
+                    }
+
+        print("Mapping out of UMLS")
+        print(tabulate(biomappings_from_umls.most_common()))
+        print("Mapping into UMLS")
+        print(tabulate(biomappings_to_umls.most_common()))
```
bgyori, line 91: An alternative approach here could be to propagate these mappings into INDRA (see https://github.com/sorgerlab/indra/blob/master/indra/resources/biomappings.tsv) and add them to the BioOntology there, then the usual standardization procedure would take care of the necessary mappings. What do you think?

bgyori:
The test failure says
```
.tox/py/lib/python3.9/site-packages/indra_cogex/sources/sider/__init__.py:17: in <module>
    from indra_cogex.constants import MODULE
E   ModuleNotFoundError: No module named 'indra_cogex.constants'
```
is a new module missing from version control?

PR #47, opened 2021-11-06
Add evidence subgraph and refactor import CLI

bgyori:
This PR makes the following changes:
- Add a new source for PubMed which creates nodes for 30M+ PubMed entries with properties including publication year and other IDs (PMCID, DOI, etc), and relations corresponding to MeSH annotations.
- Add a new source for evidences for INDRA Statements carrying all evidence data and also linked to PubMed nodes
- Refactor import CLI with a more intuitive set of flags and explicitly separating the processing, assembly, and import phases
- Separate node assembly by node type to accommodate multiple node types with distinct assembly groupings

### DIFF
--- a/src/indra_cogex/sources/cli.py
+++ b/src/indra_cogex/sources/cli.py
@@ -118,7 +118,7 @@ def main(
             processed = False
         edge_paths.append(processor_cls.edges_path)
         click.secho(
-            f"Identified node paths for assembly: {[str(p) for p in processor_to_assemble_paths]}",
+            f"Identified node paths for assembly: {[str(p) for p in processor_to_assemble_paths.values()]}",
             fg="blue",
         )
         click.secho(
--- a/src/indra_cogex/sources/cli.py
+++ b/src/indra_cogex/sources/cli.py
@@ -69,6 +69,7 @@ def main(
 ):
     """Generate and import Neo4j nodes and edges tables."""
     nodes_path = Path(nodes_path)
+    preprocessed_nodes_paths = []
     config = {} if config is None else json.load(config)
     paths = []
     na = NodeAssembler()
@@ -91,15 +92,29 @@ def main(
                     click.secho(f"Failed: {e}", fg="red")
                     continue
                 click.secho("Processing...", fg="green")
+                # First dump the nodes and edges for processor
                 _, nodes, _ = processor.dump()
+                # Only assemble BioEntity nodes
+                if processor_cls.node_type == "BioEntity":
+                    na.add_nodes(nodes)
+                else:
+                    # These nodes do not need to be assembled, we just need to
+                    # store the path to the file
+                    preprocessed_nodes_paths.append(processor_cls.nodes_path)
             else:
-                click.secho(
-                    f"Loading cached nodes from {processor_cls.nodes_indra_path}",
-                    fg="green",
-                )
-                with open(processor_cls.nodes_indra_path, "rb") as fh:
-                    nodes = pickle.load(fh)
-            na.add_nodes(nodes)
+                # Only assemble BioEntity nodes
+                if processor_cls.node_type == "BioEntity":
+                    click.secho(
+                        f"Loading cached nodes from {processor_cls.nodes_indra_path}",
+                        fg="green",
+                    )
+                    with open(processor_cls.nodes_indra_path, "rb") as fh:
+                        nodes = pickle.load(fh)
+                        na.add_nodes(nodes)
+                else:
+                    # These nodes do not need to be assembled, we just need to
+                    # store the path to the file
+                    preprocessed_nodes_paths.append(processor_cls.nodes_path)
 
         paths.append((processor_cls.nodes_path, processor_cls.edges_path))
 
--- a/src/indra_cogex/sources/cli.py
+++ b/src/indra_cogex/sources/cli.py
@@ -89,9 +89,7 @@ def main(
     """Generate and import Neo4j nodes and edges tables."""
     to_assemble = ["BioEntity", "Publication"]
     # Paths to files with preprocessed nodes (e.g. assembled nodes or nodes that don't need to be assembled)
-    nodes_paths_for_import = [
-        _get_assembled_path(node_type) for node_type in to_assemble
-    ]
+    nodes_paths_for_import = []
     config = {} if config is None else json.load(config)
     edge_paths = []
     node_assemblers = {}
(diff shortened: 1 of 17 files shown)


cthoyt:
Where do the pubmed resources come from? Can you add code that automates downloading and preprocessing them?

bgyori:
These come from custom exports from our internal resources and it would be a major undertaking to construct from scratch.

src/indra_cogex/sources/pubmed/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+                data = {
+                    "trid": get_val(trid),
+                    "pmcid": get_val(pmcid),
+                    "doi": get_val(doi),
+                    "pii": get_val(pii),
+                    "url": get_val(url),
+                    "manuscript_id": get_val(manuscript_id),
+                    "year": year,
+                }
+                yield Node("PUBMED", pmid, labels=[pmid_node_type], data=data)
```
cthoyt, line 62: maybe better to use `"ARTICLE"` in case there are ever other resources that don't have pubmed identifiers

cthoyt:
Do we need to make submodules if the only parts are `__init__.py` and `__main__.py`? Why not just make `pubmed.py`?

src/indra_cogex/sources/processor.py
```diff
@@ -88,7 +88,13 @@ def dump(self):
 
     def _dump_nodes(self) -> Path:
         sample_path = self.module.join(name="nodes_sample.tsv")
-        nodes = sorted(self.get_nodes(), key=lambda x: (x.db_ns, x.db_id))
+        nodes = sorted(
+            [
+                n
+                for n in tqdm(self.get_nodes(), desc="Node generation", unit_scale=True)
```
cthoyt, line 94: `sorted()` can take an iterable directly, so no need to do the trivial list comprehension around it

cthoyt: I made a commit to update this

src/indra_cogex/sources/pubmed/__init__.py
bgyori: No but this is specifically a prefix for a PubMed ID so it has to be PUBMED.

src/indra_cogex/sources/pubmed/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+        text_refs_path=TEXT_REFS,
+    ):
+        self.mesh_pmid_path = mesh_pmid_path
+        self.pmid_year_path = pmid_year_path
+        self.text_refs_path = text_refs_path
+
+    def get_nodes(self):
+        # We iterate over text refs to get the nodes and
+        # then look up the year to add as a property
+        pmid_node_type = "Publication"
```
bgyori, line 34: @cthoyt this is where the more generic node label is defined, not specific to PubMed.

cthoyt: oh I totally misread this. thanks for clarifying

bgyori:
> Do we need to make submodules if the only parts are `__init__.py` and `__main__.py`? Why not just make `pubmed.py`?

I am just following the pattern established early on. My preference would actually be to keep the submodules (to allow flexibly adding more files related to each source without structural changes) but to factor out the processor itself and processing-related code into a named file like `processor.py`, and then just import whatever needs to be exposed in `__init__.py`.

bgyori:
@dianakolusheva, while trying this, the realization hit again that the current CLI arguments are confusing and difficult to use. The proposed refactoring would be:
```
--process and --force-process
--assemble and --force-assemble
--import
```
for the three stages. The old `--load` `--load_only` and `--force` options would be removed.

We should also rename `--allow-missing` to `--skip-failed-processors` to make it more explicit, allow-missing is not obviously related to the logic of the code.

src/indra_cogex/sources/cli.py
```diff
@@ -17,29 +17,47 @@
...
+    nodes_path = pystow.join(
+        "indra", "cogex", "assembled", name=f"nodes_{node_type}.tsv.gz"
+    )
+    return Path(nodes_path)
+
+
 @click.command()
 @click.option(
-    "--load",
+    "-f",
```
bgyori, line 34: Remove the extra `-f` options

src/indra_cogex/sources/cli.py
```diff
@@ -53,71 +71,129 @@ def _iter_resolvers() -> Iterable[Type[Processor]]:
     " and values are dictionaries matching the __init__ parameters for the processor",
 )
 @click.option(
-    "--nodes-path",
-    default=DEFAULT_NODES_PATH,
+    "--skip_failed_processors",
+    is_flag=True,
+    help="If true, doesn't explode on missing files",
```
bgyori, line 76: If true, skips processors that are missing required input files without erroring.

src/indra_cogex/sources/cli.py
```diff
@@ -53,71 +71,129 @@ def _iter_resolvers() -> Iterable[Type[Processor]]:
...
+            fg="blue",
+        )
+        # Run the processor if needed
+        if force_process or (process and not processed):
+            try:
+                processor = processor_cls(**config.get(processor_cls.name, {}))
+            except FileNotFoundError as e:
+                if not skip_failed_processors:
+                    raise
+                click.secho(f"Failed: {e}", fg="red")
```
bgyori, line 139: More verbose: Failed: ..., skipping corresponding nodes and relations from further processing and import

src/indra_cogex/sources/cli.py
```diff
@@ -53,71 +71,129 @@ def _iter_resolvers() -> Iterable[Type[Processor]]:
...
+                        node_assemblers[node_type].add_nodes(nodes)
+        elif processed:  # force_process=False, process=True/False
+            # If we don't need to assemble, we'll just skip this
+            for node_type, nodes_indra_path in processor_to_assemble_paths.items():
+                assembled_path = _get_assembled_path(node_type)
+                if force_assemble or (assemble and not assembled_path.exists()):
+                    # Instantiate the assembler or add nodes to existing assembler
+                    if node_type not in node_assemblers:
+                        node_assemblers[node_type] = NodeAssembler()
+                        click.secho(
```
bgyori, line 165: This should be pulled out one level

src/indra_cogex/sources/cli.py
```diff
@@ -53,71 +71,129 @@ def _iter_resolvers() -> Iterable[Type[Processor]]:
...
+                    node_assemblers[node_type].add_nodes(nodes)
 
-    if not load_only:
-        if force or not nodes_path.is_file():
-            # Now create and dump the assembled nodes
-            assembled_nodes = na.assemble_nodes()
-            assembled_nodes = sorted(assembled_nodes, key=lambda x: (x.db_ns, x.db_id))
-            Processor._dump_nodes_to_path(assembled_nodes, nodes_path)
+    # Assemble nodes if we got any node assemblers above
+    if node_assemblers:
```
bgyori, line 174: We can skip the if here probably

src/indra_cogex/sources/cli.py
```diff
@@ -53,71 +71,129 @@ def _iter_resolvers() -> Iterable[Type[Processor]]:
...
+        for node_type, assembler in node_assemblers.items():
+            assembled_path = _get_assembled_path(node_type)
+            click.secho(f"Assembling {node_type}", fg="green")
+            # This path is already in the list of nodes to import
+            assembler.assemble(assembled_path)
```
bgyori, line 179: The above probably needs to be changed, I blame the copilot for this one :)

PR #49, opened 2021-11-30
Implement higher-level queries

bgyori:
This PR moves the `neo4j_client` into a dedicated `client` module and adds a new `queries` module that implements queries with higher-level semantics on top of the generic low-level ones implemented in the client. These queries are useful for analysis workflows and human interaction.

### DIFF
--- a/src/indra_cogex/client/subnetwork.py
+++ b/src/indra_cogex/client/subnetwork.py
@@ -1,0 +1,56 @@
+from typing import List, Tuple
+from .queries import get_expressed_genes_in_tissue
+from .neo4j_client import Neo4jClient
+from ..representation import norm_id, Node
+
+
+def indra_subnetwork(client: Neo4jClient, nodes: List[Tuple[str, str]]) -> List[Node]:
+    """Return the INDRA Statement subnetwork induced by the given nodes.
+
+    Parameters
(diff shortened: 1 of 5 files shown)


src/indra_cogex/client/subnetwork.py
```diff
@@ -0,0 +1,56 @@
+from typing import List, Tuple
+from .queries import get_expressed_genes_in_tissue
+from .neo4j_client import Neo4jClient
+from ..representation import norm_id, Node
+
+
+def indra_subnetwork(client: Neo4jClient, nodes: List[Tuple[str, str]]) -> List[Node]:
```
cthoyt, line 7: is the return type annotation right?

src/indra_cogex/representation.py
```diff
@@ -166,3 +168,23 @@ def norm_id(db_ns, db_id):
...
+    ----------
+    rels :
+        A list of Relations.
+
+    Returns
+    -------
+    :
+        A list of INDRA Statements.
+    """
+    stmts_json = [json.loads(rel.data["stmt_json"]) for rel in rels]
```
cthoyt, line 188: maybe just bake a conversion function into the Relation class?

PR #50, opened 2022-01-05
MeSH-PubMed extension of mid level queries

kkaris:
This PR extends the number of mid-level queries available in the `queries` sub-module of the `client` module.

Functions added:
- `get_pmids_for_mesh_id`
- `get_mesh_ids_for_pmid`
- `get_evidence_obj_for_stmt_hash`
- `get_evidence_objects_for_stmt_hashes`
- `get_stmts_for_pmid`
- `get_stmts_for_mesh_id`
- `get_stmts_for_stmt_hashes`

### DIFF
--- a/src/indra_cogex/client/queries.py
+++ b/src/indra_cogex/client/queries.py
@@ -658,7 +658,9 @@ def get_stmts_for_pmid(
     return indra_stmts_from_relations(rels)
 
 
-def get_stmts_for_mesh_id(client: Neo4jClient, meshid: Tuple[str, str]) -> Iterable[Statement]:
+def get_stmts_for_mesh_id(
+    client: Neo4jClient, meshid: Tuple[str, str], include_child_terms: bool = True
+) -> Iterable[Statement]:
     """Return the statements with evidence for the given MESH ID.
 
     Parameters
@@ -667,6 +669,8 @@ def get_stmts_for_mesh_id(client: Neo4jClient, meshid: Tuple[str, str]) -> Itera
         The Neo4j client.
     meshid :
         The MESH ID to query.
+    include_child_terms :
+        If True, also match against the children of the given MESH id.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


src/indra_cogex/client/queries.py
```diff
@@ -415,3 +424,273 @@ def isa_or_partof(
...
+    if mesh[0].lower() != "mesh":
+        raise ValueError(f"Expected mesh term, got {':'.join(mesh)}")
+
+    # NOTE: we could use get_ontology_child_terms() here, but it's ~20 times
+    # slower for this specific query, so we do an optimized query that is
+    # basically equivalent instead.
+    if include_child_terms:
+        child_query = (
+            """
+            MATCH (c:BioEntity)-[:isa|partf*1..]->(p:BioEntity)
```
cthoyt, line 461: Typo in part of?

bgyori: we don't need partof here

src/indra_cogex/client/queries.py
```diff
@@ -415,3 +424,273 @@ def isa_or_partof(
...
+            WHERE p.id = "mesh:%s"
```
bgyori, line 462: You can remove this constraint

src/indra_cogex/client/queries.py
kkaris: Yup, will fix and check query

kkaris: Although, I just realized, isn't the mesh hierarchy strictly `isa` relationships? Then I shouldn't use `partof` in the first place.

kkaris: Yes:
https://github.com/sorgerlab/indra/blob/master/indra/ontology/bio/ontology.py#L368-L370

kkaris: Resolved in https://github.com/bgyori/indra_cogex/pull/50/commits/e9c784e5ae34e55686b5e77e273bfc47d474c09e

kkaris: (oh, sorry, your comment wasn't visible until now @bgyori, the page must have been outdated when it woke up the laptop after coming in)

kkaris:
One thing that likely will speed up the queries that return indra statements is to try to skip querying for evidence twice. Right now in `get_stmts_for_pmid()` it goes pmid -> evidence -> stmt hash -> statement ->Evidence. That was shortsighted coding on my part, since I made functions for each sub query. I working on a fix that'll hopefully speed things up.

tests/test_queries.py
```diff
@@ -0,0 +1,183 @@
...
+    assert node0.db_ns == 'GO'
+
+
+@pytest.mark.nonpublic
+def test_get_genes_for_go_term():
+    # Single query
+    client = _get_client()
+    go_term = ("GO", "GO:0000978")
+    genes = get_genes_for_go_term(client, go_term)
+    assert len(genes) == 1146
```
cthoyt, line 56: I'd be careful to hard-code numbers in this (and all of the other) tests here, since these might change if the GO Annotations database or other underlying data sources get updated. It's much less likely that a single element will be removed, so perhaps check for something specific?

kkaris: Good point, will fix

kkaris:
Branch tests should be running now on my branch: https://github.com/kkaris/indra_cogex/actions

src/indra_cogex/client/queries.py
```diff
@@ -840,6 +850,32 @@ def get_stmts_for_stmt_hashes(
     return list(stmts.values())
 
 
+def _get_mesh_child_terms(client: Neo4jClient, meshid: Tuple[str, str]) -> Set[str]:
```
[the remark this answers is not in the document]
kkaris: Or did I misunderstand? Did you propose to rename all the id inputs like this `<type>_id` -> `<type>_curie`?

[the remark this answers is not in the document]
cthoyt: At minimum I would suggest `<type>_id` is better than `<type>id`.

Maybe `<type>_curie` is more descriptive than `<type>_id`, but on second thought, this would mean we would have to change other code too.

bgyori (approved):

PR #52, opened 2022-01-06
Add optional source and target constraints to some queries

cthoyt:
This should improve the performance of `get_genes_for_go_term()`

- [x] #53
- [x] Add additional places for constraints
- [ ] Testing (borrow tests from #50?)

This PR also adds a very tiny configuration for `isort` to keep import order consistent (in case pycharm reorganizes them, it's a pain to undo without isort)

PR #53, opened 2022-01-06
Implement DSL for Cypher triple queries

cthoyt:
While working on #52 it occurred to me that I would have to non-trivially re-write a lot of these match statements to add in type constraints. This PR adds a more generic function that is amenable to easily adding type constraints. This PR has associated unit tests that covers the 5 Cypher statements that it replaces.

Caveat: it doesn't have a first class notion of the `*1...` that appears in the predecessor/successor nodes. I could add that, but it might be premature generalization, so I left a TODO for later


bgyori (approved):
Looks good, thanks! I'm not sure representation.py is a perfect fit for placement but it should be fine for now.

PR #55, opened 2022-01-14
Add indexing client and module

kkaris:
This PR does 3 updates:

1. Add node and relationship indexing methods to the neo4j client that creates a single-property b-tree index for [nodes](https://neo4j.com/docs/cypher-manual/4.4/indexes-for-search-performance/#administration-indexes-create-a-single-property-b-tree-index-for-nodes) or [relationships](https://neo4j.com/docs/cypher-manual/4.4/indexes-for-search-performance/#administration-indexes-create-a-single-property-b-tree-index-for-relationships), respectively. _Note that relationship indexing is only available in neo4j 4.3+ (see [release notes](https://neo4j.com/release-notes/database/neo4j-4-3-0/) and [blog post](https://medium.com/neo4j/neo4j-4-3-blog-series-relationship-indexes-14d2b656abd2))_
2. Add CLI commands with `click` for building indexes.
3. Update field types for some node and relationship properties from `string` to e.g. `int` or `boolean`. This will allow a cypher query to filter on properties according to their field type, e.g.

```
MATCH (k:Publication)-[:annotated_with]->(:BioEntity)
WHERE k.year > 2010
RETURN DISTINCT k
```

### DIFF
--- a/src/indra_cogex/indexing/cli.py
+++ b/src/indra_cogex/indexing/cli.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 """Run the database indexing."""
+from typing import Optional, Tuple
 
 import click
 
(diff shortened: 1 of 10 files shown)


src/indra_cogex/indexing/cli.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+"""Run the database indexing."""
+from typing import Optional, Tuple
+
+import click
+
+
+@click.command(name="Build extra indexes")
```
cthoyt, line 9: this name  might not do what you expect - this would get used if it's added as a subcommand in another CLI (spaces are an issue)

src/indra_cogex/sources/pubmed/__init__.py
```diff
@@ -89,7 +89,7 @@ def get_relations(self):
                             "MESH",
                             mesh_id,
                             "annotated_with",
-                            {"is_major_topic": True if major_topic == "1" else False},
```
cthoyt, line 92: 🚀

src/indra_cogex/indexing/cli.py
kkaris: Ah, beginner's mistake, will fix

kkaris: Resolved in https://github.com/bgyori/indra_cogex/pull/55/commits/89cdb5a00b64636b41e705992e6019c30c44262d

src/indra_cogex/sources/pubmed/__init__.py
kkaris: Sorry @cthoyt, had to extend the expression again after reading the[nedo4j admin manual](https://neo4j.com/docs/operations-manual/current/tools/neo4j-admin/neo4j-admin-import/#import-tool-header-format-properties) again carefully:

> Boolean values are true if they match exactly the text true. All other values are false.

Fixed in https://github.com/bgyori/indra_cogex/pull/55/commits/7b40c913efcc8308229beb69fedb0e5f6a8e138c

bgyori (approved):
Nice!

PR #57, opened 2022-01-22
Fix query and images for gene list analysis

bgyori:
This PR fixes one of the queries for analysis. Since the downstream/upstream images were overflowing their container, I tried decreasing their size but this made them flow into the results list. @cthoyt if you can figure out how to avoid this, please push to this branch.


setup.cfg
```diff
@@ -64,7 +64,7 @@ tests =
 web =
     flask
     flask-wtf
-    bootstrap-flask
+    bootstrap-flask<=2.0.0
```
bgyori, line 67: We might want to add a lower bound as well depending on when Bootstrap4 was added - this didn't work with my 1.7.0 version.

bgyori: Or did you actually mean to put `>=` here?

bgyori: I concluded that the only logical explanation is that it should be >=

cthoyt: Thanks for the typo spot, you were correct

PR #58, opened 2022-01-25
Add database summary to GLA app

cthoyt:
This PR creates a page that summarizes the graph database's contents.

I'm not sure if I want to add everything into the gene list analysis application, but for now it was more convenient since I didn't want to duplicate some of the base functionality. A middle ground would be to use different flask Blueprints for parts of the app that could be imported/run depending on the context.

## Demo

It looks like this (the numbers tick up from zero, which is kind of fun)

Screen Shot 2022-01-25 at 17 53 01
.


bgyori:
That looks nice, I think we could also add a static version of the image to the README directly.

cthoyt:
@bgyori a high quality summary image is now in the readme

bgyori (approved):
Nice, thanks!

PR #61, opened 2022-01-28
Add significance cutoff button to analysis forms

cthoyt:
Add a button to each of the discrete, signed, and continuous analysis forms to opt-in to keeping insignificant results.


bgyori (approved):
Great, thanks!

PR #63, opened 2022-01-28
More queries

dianakolusheva:
This PR adds a few more high level queries.

### DIFF
--- a/src/indra_cogex/client/queries.py
+++ b/src/indra_cogex/client/queries.py
@@ -7,7 +7,7 @@ import networkx as nx
 from indra.statements import Evidence, Statement, stmts_from_json, Agent
 
 from .neo4j_client import Neo4jClient, autoclient
-from ..representation import Node,Relation,  indra_stmts_from_relations, norm_id
+from ..representation import Node, Relation, indra_stmts_from_relations, norm_id
 
 logger = logging.getLogger(__name__)
 
@@ -398,6 +398,7 @@ def get_shared_pathways_for_genes(
     )
 
 
+@autoclient()
 def get_genes_for_pathway(
     pathway: Tuple[str, str], *, client: Neo4jClient
 ) -> Iterable[Node]:
@@ -1031,6 +1032,7 @@ def get_schema_graph(*, client: Neo4jClient) -> nx.MultiDiGraph:
 # CCLE
 
 
+@autoclient()
 def is_gene_mutated(
     gene: Tuple[str, str], cell_line: Tuple[str, str], *, client: Neo4jClient
 ) -> bool:
@@ -1062,6 +1064,7 @@ def is_gene_mutated(
 # Indra DB
 
 
+@autoclient()
 def get_drugs_for_target(
     target: Tuple[str, str], *, client: Neo4jClient
 ) -> Iterable[Agent]:
@@ -1076,6 +1079,7 @@ def get_drugs_for_target(
     return drug_nodes
 
 
+@autoclient()
 def get_targets_for_drug(
     drug: Tuple[str, str], *, client: Neo4jClient
 ) -> Iterable[Agent]:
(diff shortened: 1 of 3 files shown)


src/indra_cogex/client/queries.py
```diff
@@ -346,6 +351,31 @@ def get_pathways_for_gene(client: Neo4jClient, gene: Tuple[str, str]) -> Iterabl
     )
 
 
+def get_shared_pathways_for_genes(
+    client: Neo4jClient, genes: Iterable[Tuple[str, str]]
```
cthoyt, line 355: can you please make the client keyword only for all new queries

cthoyt: see the autoclient decorator branch #60 which will need this

src/indra_cogex/client/queries.py
```diff
@@ -952,3 +982,80 @@ def get_schema_graph(client: Neo4jClient) -> nx.MultiDiGraph:
...
+
+
+def is_drug_target(
+    client: Neo4jClient, drug: Tuple[str, str], target: Tuple[str, str]
+) -> bool:
+    """Return True if the drug targets the given protein."""
+    rels = client.get_relations(
+        drug, target, "indra_rel", source_type="BioEntity", target_type="BioEntity"
+    )
+    return any(_is_drug_relation(rel) for rel in rels)
```
cthoyt, line 1056: could be potentially faster to bake the statement type filter into the query itself

src/indra_cogex/client/queries.py
dianakolusheva: Done

src/indra_cogex/client/queries.py
```diff
@@ -952,3 +982,82 @@ def get_schema_graph(client: Neo4jClient) -> nx.MultiDiGraph:
...
+    """Return the drugs targetting the given protein."""
+    rels = client.get_source_relations(
+        target, "indra_rel", source_type="BioEntity", target_type="BioEntity"
+    )
+    drug_rels = [rel for rel in rels if _is_drug_relation(rel)]
+    drug_nodes = [(rel.source_ns, rel.source_id) for rel in drug_rels]
+    stmt_jsons = [json.loads(rel.data["stmt_json"]) for rel in drug_rels]
+    stmts = stmts_from_json(stmt_jsons)
+    drug_agents = [stmt.subj for stmt in stmts]
+    return {"nodes": drug_nodes, "agents": drug_agents}
```
bgyori, line 1033: I find it a little bit odd that both nodes and agents are returned here. If the nodes are provided, they ought to represent the grounding of the drug such that 
```
stmt_jsons = [json.loads(rel.data["stmt_json"]) for rel in drug_rels]
stmts = stmts_from_json(stmt_jsons)
drug_agents = [stmt.subj for stmt in stmts]
```
should be redundant, I think, and could be replaced by outside code doing
```
get_standard_agent(node.data.name, {node.db_ns: node.db_id})
```
if constructing an Agent is needed.

src/indra_cogex/client/queries.py
bgyori: I'm not sure that's possible with the current API of get_relations (since we don't have a slot for specifying edge property constraints). Since both drug and target are given here, I don't think it will make a big difference.

src/indra_cogex/client/queries.py
dianakolusheva: I agree that it makes more sense to return the nodes only at the query level. Two reasons why I implemented it this way here:
1) since we have to filter based on relation properties (Inhibition from "tas") and not on the node properties, we get relations back, so instead of full node objects we get relation.source_ns and relation.source_id (so no name and other node data_
2) agents are available here since we have statement json

I think the solution here would be to get access to the full node data when we only have a ns, id pair

PR #64, opened 2022-01-30
Enable analysis on human and mouse data in GSEA app

cthoyt:
*note not tested yet


bgyori (approved):

cthoyt:
Same situation with this - tests are still passing locally but there's an issue with a sigkill on CI.

PR #68, opened 2022-02-02
Fix typo in method name in subnetwork module

cthoyt:


cthoyt:
@bgyori this was really uncontroversial so I went ahead and merged it so I could reuse this in another PR

PR #69, opened 2022-02-02
Metabolite set enrichment analysis

### LINKED ISSUE #67: Metabolite enrichment analysis
In addition to the gene set enrichment analysis, it makes sense to be able to do metabolite enrichment analysis. Though there are lots of possible chemical information databases for that contain metabolites (ChEBI, HMDB, MeSH, ChEMBL), it's probably best to start just by intaking ChEBI identifiers.

Potential queries:

1. Protein increases metabolite, return protein
2. Protein in family, protein increases metabolite, return family

## Curation and Example Generation Workflow

1. Input EC-Code
2. Look up Rhea reaction
3. Parse product side of Rhea reaction
4. Filter out several common things from products:
   - NAD+
   - NADH
   - H+
   - H20
   - OH-
5. Use remaining products as "filter" for high-quality statements using RDKit substructure matching

## Potential Examples

[FPLX:ALDH](https://bioregistry.io/FPLX:ALDH) or [ECCODE:1.2.1.3](https://bioregistry.io/ECCODE:1.2.1.3) has the reaction `aldehyde + NAD+ + H2O = a carboxylate + NADH + H+` and corresponds to [rhea:16185](https://bioregistry.io/rhea:16185).

cthoyt:
Closes #67 

This PR is the first step towards metabolism set enrichment analysis. It does the following:

1. Makes a new query for EC - FPLX - metabolite
2. Implements ORA based on this query
3. Massive reorganization of webapp to make extensible to new query types
4. Add Metabolite set enrichment analysis
5. Adds a switch with the `INDRA_COGEX_WEB_LOCAL` constant to turn on the local download box (this is potentially risky to create files on a machine hosting this service, so it's really not standard. This was for Sam's use case last week)

**Caveat** at the moment, this PR treats "metabolite" as synonymous with "ChEBI entry". This is obviously not correct due to the limited scope of ChEBI. Future work could go into generalizing this to all multiple namespaces, then again narrowing it to only molecules that are "metabolites" (which itself is a difficult philosophical question)

## How to review:

1. Spin up the web application with `python -m indra_cogex.apps.gla -vv`
2. You'll see the MSEA analysis tab. Have fun!


cthoyt:
@bgyori this is ready, I don't think it needs another review since we already went through this and nothing has changed since besides rebasing on other commits

kkaris:
I tested the app locally and the got tsv files to download from `/gene/discrete`, but it coincides with a redirect to `/gla.discretize_analysis` that doesn't work. @cthoyt did you mean to redirect back to `/gene/discretize`?

kkaris:
I see, I fixed the hacked solution to downloads that we implemented for Sam, so not urgent then :)

bgyori (approved):

PR #70, opened 2022-02-09
Queries web app

kkaris:
This PR adds a simple flask app wrapper around all query functions in the query module.

In the future this Flask app could be changed to a FastAPI app to accomodate `async` handling, automatic type checking with `pydantic` and `GraphQL` [support](https://fastapi.tiangolo.com/advanced/graphql/) if we want. This app could also simply be mounted (see [here](https://fastapi.tiangolo.com/advanced/wsgi/?h=flask) and [here](https://fastapi.tiangolo.com/advanced/sub-applications/)) as a sub-app to a larger web api.

### DIFF
--- a/src/indra_cogex/apps/query_web_app/__init__.py
+++ b/src/indra_cogex/apps/query_web_app/__init__.py
@@ -315,8 +315,8 @@ def evidences_for_mesh():
     mesh_term_name = request.json.get("mesh_term_name")
     if mesh_term_name is None:
         abort(Response("Parameter 'mesh_term_name' not provided", 415))
-    evidences = get_evidences_for_mesh(client, mesh_term_name)
-    return jsonify({h: [e.to_json() for e in el] for h, el in evidences.items()})
+    evidence_dict = get_evidences_for_mesh(client, mesh_term_name)
+    return jsonify({h: [e.to_json() for e in el] for h, el in evidence_dict.items()})
 
 
 @app.route("/get_evidences_for_stmt_hash", methods=["POST"])
@@ -341,8 +341,8 @@ def evidences_for_stmt_hashes():
     stmt_hashes = request.json.get("stmt_hashes")
     if stmt_hashes is None:
         abort(Response("Parameter 'stmt_hashes' not provided", 415))
-    evidences = get_evidences_for_stmt_hashes(client, stmt_hashes)
-    return jsonify({h: [e.to_json() for e in el] for h, el in evidences.items()})
+    evidence_dict = get_evidences_for_stmt_hashes(client, stmt_hashes)
+    return jsonify({h: [e.to_json() for e in el] for h, el in evidence_dict.items()})
 
 
 @app.route("/get_stmts_for_pmid", methods=["POST"])
--- a/src/indra_cogex/apps/query_web_app/__init__.py
+++ b/src/indra_cogex/apps/query_web_app/__init__.py
@@ -40,13 +40,8 @@ def _post_request_preproc(*keys) -> Union[TupOfTups, Tup]:
 @app.route("/get_genes_in_tissue", methods=["POST"])
 def genes_in_tissue():
     """Get genes for a disease."""
-    if request.json is None:
-        abort(Response("Missing application/json header.", 415))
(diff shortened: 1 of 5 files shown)


src/indra_cogex/apps/query_web_app/__init__.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def _post_request_preproc(*keys) -> Union[TupOfTups, Tup, str, int]:
+    if request.json is None:
+        abort(Response("Missing application/json header.", 415))
+    tups = []
+    for key in keys:
+        if key not in request.json:
+            abort(Response("Parameter '%s' not provided" % key, 415))
+        tups.append(tuple(request.json[key]))
```
cthoyt, line 32: maybe cleaner to use `request.args.get()`? I think this exposes the path and parameters together

kkaris: Good point, definitely worth looking into to reduce code replication.

kkaris: Although, if using an endpoint like `/<query_func>` to match all available functions, then we can't have the REST API docs for each endpoint listed..

kkaris: I tried providing a set of allowed values for a `/{func_name}` type endpoint in FastAPI, but that didn't get pretty
image

cthoyt: you could potentially use a for loop to programatically generate different endpoints

kkaris: Something like this
```python
from fastapi import FastAPI
from typing import Dict, List
from indra_cogex.client import queries


app = FastAPI()


for query_func in queries.__all__:
    @app.post(f"/{query_func}", response_model=List[Dict])
    async def f(term: str):
        return [{"term": term}]
```
making this:
image

The more verbose part of this would be to make each endpoint unique to its the parameters and theis types as well as the return type.

cthoyt: time to break out the `inspect` library!

kkaris:
Ready for final review

bgyori:
This is really cool! I'm trying it out and will comment if I run into any issues.

kkaris:
Just realized all the returns all have the format`name : <type annotation>`, I'll fix that

src/indra_cogex/apps/query_web_app/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+                raise ValueError(f"{key} must be a string or list of strings")
+        else:
+            parsed_query[key] = value
+
+    return parsed_query
+
+
+def process_result(result) -> Any:
+    # Any fundamental type
+    if isinstance(result, (int, str, bool)):
```
cthoyt, line 89: float?

src/indra_cogex/apps/query_web_app/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+        return result
+    # Any iterable query
+    elif isinstance(result, (Iterable, list, set)):
```
cthoyt, line 92: doing an instance check on Iterable might get you in trouble if it's a dict on certain versions of python, so I'd suggest switching the order of the checks so the dict one comes first

src/indra_cogex/apps/query_web_app/__init__.py
kkaris: Fixed in https://github.com/bgyori/indra_cogex/pull/70/commits/bfad9aba3bc0cfb785c874e02753d4955460df61

src/indra_cogex/apps/query_web_app/__init__.py
kkaris: Fixed in https://github.com/bgyori/indra_cogex/pull/70/commits/ba71e2576a6766ac76d98a44ebacbe933a443ff7

cthoyt (approved):

bgyori (approved):

PR #74, opened 2022-02-22
Fix query tests

kkaris:
This PR updates the query functions in `queries.py` and the query tests in `test_queries.py` to:

- Adapt to that some node and edge properties have switched their field types from string to something else
- Speed up some queries by setting the node type

### DIFF
--- a/src/indra_cogex/client/queries.py
+++ b/src/indra_cogex/client/queries.py
@@ -740,7 +740,7 @@ def get_evidences_for_mesh(
 
 @autoclient()
 def get_evidences_for_stmt_hash(
-    stmt_hash: Union[str, int], *, client: Neo4jClient
+    stmt_hash: int, *, client: Neo4jClient
 ) -> Iterable[Evidence]:
     """Return the matching evidence objects for the given statement hash.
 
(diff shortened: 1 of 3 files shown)


src/indra_cogex/client/queries.py
```diff
@@ -784,7 +794,7 @@ def get_evidences_for_stmt_hashes(
         A mapping of stmt hash to a list of evidence objects for the given
         statement hashes.
     """
-    stmt_hashes_str = ",".join(f'"{h}"' for h in stmt_hashes)
+    stmt_hashes_str = ",".join(f"{h}" for h in stmt_hashes)
```
cthoyt, line 797: just `str(h)`?

tests/test_queries.py
```diff
@@ -290,24 +291,24 @@ def test_get_stmts_for_pmid():
 def test_get_stmts_for_mesh_id_w_children():
     # Two queries:
     # 1. evidences for publications with pmid having mesh annotation
-    # 2. statements for the evidences in 2
+    # 2. statements for the evidences in 1
     client = _get_client()
     mesh_id = ("MESH", "D000068236")
     stmts = get_stmts_for_mesh(mesh_id, client=client)
     assert stmts
-    assert isinstance(stmts[0], Activation)
```
cthoyt, line 298: why did this change?

src/indra_cogex/client/queries.py
kkaris: I guess, it's the same number of characters :)

tests/test_queries.py
kkaris: Not sure, I checked the statements from the query and the Inhibition is new since before as I remember it. Anyway, it's probably a bad idea to test something that's not static.

bgyori (approved):

PR #75, opened 2022-02-23
Implement queries for relations for multiple sources and targets

bgyori:
This PR implements new generic queries for relations from/to multiple sources/targets, and implements high-level queries on top of these to allow querying for the targets of multiple drugs and drugs for multiple targets in a single query.


cthoyt:
@bgyori this is great! I was thinking about this yesterday with collect statements but ultimately didn't like my solution since it wasn't very reusable. I hope you don't mind but I'm going to try this out and push some test cases

PR #82, opened 2022-03-22
Add first curation workflows

cthoyt:
This PR does the following:

1. Creates a dedicated curation blueprint (moves the GO curation bits out of the data_display blueprint)
2. Implements prototypes of new curation workflows
   1. Pathway Curator   
   2. MeSH term-driven curation workflow
   3. GO Annotation Curator
   4. Curation conflict resolution workflow
   5. PPI Curator
   6. Transcription Factor Cuator
   7. Kinase Curator
   8. Phosphatase Curator
   9. DUB curator
3. Implements first-pass user interfaces for beginning curation
4. Adds links to home page

All of the queries take less than a minute to generate the curation pages. on a first pass, I think this is good enough, since the user will typically spend a lot of time on the page after it's been loaded

# TODO List Before Merge

- [x] Look one more time into ordering
- [x] benchmark times for all apps
- [x] consider cutting out some more basic changes into separate PR


bgyori:
Should we switch this to "Ready for review"?

cthoyt:
@bgyori I will go through my checklist monday morning then we can review it on Monday if you've got time

bgyori (approved):
Super cool! 🎉

PR #88, opened 2022-03-30
Make extra evidence loadable

kkaris:
This PR updates the statement display base template to pass in `:loadable="true"` to the `<statement>` component. This turns on the "load more" functionality.

Note that the initial requests from the graph database come with all the evidence for the statements, so this functionality currently just shows evidence that's already present on the loaded page.

As a followup, we might want to experiment with _not_ querying for evidence on the first render but then allow for a follow up-query of the evidence once it's clicked and see if that gives a better user experience (fast page load but the some wait time when expanding the list of evidences)

_edit: rebased on top of_ #87


src/indra_cogex/client/queries.py
```diff
@@ -787,8 +800,14 @@ def get_evidences_for_stmt_hash(
                RETURN n.evidence"""
         % stmt_hash
     )
+
+    # Add limit and offset
+    if offset > 0:
```
cthoyt, line 805: if you add an offset/limit you also need to add an `ORDER BY` to ensure the queries are consistent

kkaris: Fixed in https://github.com/bgyori/indra_cogex/pull/88/commits/416ae8cdb4514232ab914d31d4000d173509722b

src/indra_cogex/client/queries.py
```diff
@@ -909,6 +950,37 @@ def get_stmts_for_mesh(
...
+
+    Returns
+    -------
+    :
+        A dict of statements with their metadata
+    """
+    stmt_hashes_str = ",".join(str(h) for h in stmt_hashes)
+    query = dedent(
+        f"""
+        MATCH p=(a)-[r:indra_rel]->(b)
```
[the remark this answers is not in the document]
kkaris: resolved in https://github.com/bgyori/indra_cogex/pull/88/commits/ee09ba9efcebc4407c98ef1cb2c83e01584163f9

bgyori (approved):

PR #92, opened 2022-04-12
Curation cache

kkaris:
This PR adds a new class that acts as an interface for the curation data base and a cache for the curations. If the cached curations are older than a set time delta, they are refreshed. Curations are also set to refresh on the first call to get curations after a curation.

### DIFF
--- a/src/indra_cogex/apps/curator/curation_cache.py
+++ b/src/indra_cogex/apps/curator/curation_cache.py
@@ -1,0 +1,118 @@
+"""
+A class to manage curations and the curation cache
+
+The curation cache should handle:
+    - Loading curations to cache in memory
+    - Submitting curations to the curation database
+    - Update the in memory cache when curations are submitted or at regular
+      intervals
+"""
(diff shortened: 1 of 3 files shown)


src/indra_cogex/apps/curator/curation_cache.py
```diff
@@ -0,0 +1,119 @@
...
+"""
+from datetime import timedelta, datetime
+from typing import List, Dict, Union, Optional
+from indra.sources.indra_db_rest import get_curations, submit_curation
+import pandas as pd
+
+
+Curations = List[Dict[str, Optional[Union[str, int]]]]
+# Store curations in a pandas dataframe
+CurationDict: pd.DataFrame
```
bgyori, line 18: Is this definition used somewhere?

kkaris: No it's a leftover. Will delete.

bgyori (approved):

PR #96, opened 2022-04-13
Add additional flags as a follow-up to #91

cthoyt:
I didn't notice there was branching logic depending on if the JSON was to be uploaded when I did #91, so this adds the same flags to the other branch.

This PR also parses the source API out of the evidence and puts it as a top-level entry in the Evidence nodes.

After this is merged and the database is rebuilt, I'll make a PR for https://github.com/bgyori/indra_cogex/compare/update-curator-queries, which updates the queries to rely on these fields.

### DIFF
--- a/src/indra_cogex/sources/indra_db/__init__.py
+++ b/src/indra_cogex/sources/indra_db/__init__.py
@@ -21,7 +21,11 @@ import pystow
 from indra.databases.identifiers import ensure_prefix_if_needed
 from indra.ontology.bio import bio_ontology
 from indra.util import batch_iter
-from indra.util.statement_presentation import db_sources, reader_sources
+from indra.util.statement_presentation import (
+    db_sources,
+    internal_source_mappings,
+    reader_sources,
+)
 from more_click import verbose_option
 from tqdm import tqdm
 
@@ -367,6 +371,9 @@ class EvidenceProcessor(Processor):
                                 evidence["pmid"],
                                 ["Publication"],
                             )
+
+                    source_api = evidence["source_api"]
+                    source_api = internal_source_mappings.get(source_api, source_api)
                     node_batch.append(
                         Node(
                             "indra_evidence",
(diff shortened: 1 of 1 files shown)


src/indra_cogex/sources/indra_db/__init__.py
```diff
@@ -366,6 +375,7 @@ def get_nodes(self) -> Iterable[Node]:
                             {
                                 "evidence:string": json.dumps(evidence),
                                 "stmt_hash:long": stmt_hash,
+                                "source_api:str": evidence["source_api"],
```
bgyori, line 378: Once complication with this is that `evidence["source_api"]` will, in this setting, not always match the source counts keys due to the fact that INDRA's input sources use nomenclature not always consistent with the INDRA DB's. Given that we are using INDRA DB nomenclature elsewhere for source metadata, we should probably map to those here.

cthoyt: So the motivation behind this one is to enable medscan filtering, so we could either ask @kkaris to make this addition (since he's the keeper of the ultimate source count map) or we could alternatively make it a "is_medscan:bool" flag instead

src/indra_cogex/sources/indra_db/__init__.py
```diff
@@ -358,6 +371,9 @@ def get_nodes(self) -> Iterable[Node]:
                                 evidence["pmid"],
                                 ["Publication"],
                             )
+
+                    source_api = evidence["source_api"]
+                    source_api = internal_source_mappings.get(source_api, source_api)
```
bgyori, line 376: @cthoyt do you remember if/why this mapping was necessary? I would normally not do this here but I might be missing something.

cthoyt: @kkaris is the master of the renaming, this was his suggestion to make sure that the names were consistent

bgyori: I looked into it a bit more deeply and with the latest approach to getting this content, this mapping shouldn't be necessary.

bgyori:
All of these flags as integrated on another branch now independently so we can close this.

PR #98, opened 2022-04-15
Add disease-phenotype and phenotype-gene processors from HPO

cthoyt:
This PR does the following:

1. Adds a processor for disease-phenotype relationships from HPO. Since HPO lists diseases with OMIM, Orphanet, and DECIPHER identifiers, it uses a combination of PyOBO and the INDRA Bioontology to standardize them.
2. Adds a processor for phenotype-gene relationships from HPO. Note that these are derived from disease-gene relationships, but still pretty useful.
3. Extends the backend for both discrete and continous gene-set enrichment analysis and make them available in the web app

Some things to discuss:

1. What should the relationships be called for these two?
2. Would it make sense to extend the Bioontology to handle OMIM and MONDO mappings to take care of this better?

### DIFF
--- a/src/indra_cogex/client/enrichment/utils.py
+++ b/src/indra_cogex/client/enrichment/utils.py
@@ -264,7 +264,7 @@ def get_reactome(*, client: Neo4jClient) -> Dict[Tuple[str, str], Set[str]]:
     return collect_gene_sets(client=client, query=query, cache_file=cache_file)
 
 
-@autoclient(cache=True)
+@autoclient()
 def get_phenotype_gene_sets(*, client: Neo4jClient) -> Dict[Tuple[str, str], Set[str]]:
     """Get HPO phenotype gene sets.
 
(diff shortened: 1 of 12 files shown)


src/indra_cogex/client/enrichment/utils.py
```diff
@@ -263,6 +264,32 @@ def get_reactome(*, client: Neo4jClient) -> Dict[Tuple[str, str], Set[str]]:
...
+    """
+    query = dedent(
+        """\
+        MATCH (s:BioEntity)-[:phenotype_has_gene]-(gene:BioEntity)
+        WHERE s.id STARTS WITH "HP" and gene.id STARTS WITH "hgnc"
+        RETURN s.id, s.name, collect(gene.id);
+    """
+    )
+    logger.info("caching phenotype gene sets with Cypher query: %s", query)
+    return collect_gene_sets(client=client, query=query)
```
kkaris, line 290: We need to either add `cache_file` here or handle when `cache_file` is `None` in `collect_gene_sets`. As it is written now, the code in collect_gene_sets will error if `cache_file=None`.

kkaris:
The two tests I see failing, test_get_evidence_obj_for_stmt_hash and test_get_evidence_obj_for_stmt_hashes, are fixed in #109 here https://github.com/bgyori/indra_cogex/pull/109/commits/8f5c8dc45f8d06bb2bf6c959237312b79f204d67 and here https://github.com/bgyori/indra_cogex/pull/109/commits/2e891e40ff3d82566190cd6e74574f15d91f8c4f.

src/indra_cogex/client/enrichment/utils.py
cthoyt: Done in c1c68a8

cthoyt: I also removed the default none so you can't even call this function without it

PR #99, opened 2022-04-20
Statement page fix

kkaris:
This PR updates the order for the sources so that the source badges in the statements view have the databases first/to the left and readers second/to the right.

The sources shown at the top are also dependent on which sources are actually present in the loaded statements.


kkaris:
The view after the PR:
image

kkaris:
Fixed:
image

PR #100, opened 2022-04-21
Medscan fix

kkaris:
This PR updates the handling of requests to `/statement_display` so that medscan is removed when necessary.

Two other updates in this PR:

- Two for loops are merged to one, which should make things more efficient
- The curation badge at the statement level now has a location specified which previously made the rendering bad

Test locally by loading statements that have some medscan, e.g. `-22470740286509891`, or only medscan (e.g. `-22478113093459888`)

### DIFF
--- a/src/indra_cogex/apps/data_display/__init__.py
+++ b/src/indra_cogex/apps/data_display/__init__.py
@@ -220,6 +220,7 @@ def get_evidence(stmt_hash):
 def statement_display():
     user, roles = resolve_auth(dict(request.args))
     email = user.email if user else ""
+    remove_medscan = user is None
 
     # Get the statements hash from the query string
     try:
@@ -262,9 +263,15 @@ def statement_display():
 
         available_sources_dict = {}
         for src_type, sources in sources_dict.items():
-            available_sources_dict[src_type] = [
-                src for src in sources if src in available_sources
-            ]
+            available_sources_dict[src_type] = []
+            for source in sources:
+                if source in available_sources:
+                    # If not logged in, skip medscan
+                    if remove_medscan and source == "medscan":
+                        continue
+
+                    available_sources_dict[src_type].append(source)
+
         return render_template(
             "data_display/data_display_base.html",
             stmts=stmts,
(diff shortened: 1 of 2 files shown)


src/indra_cogex/apps/data_display/__init__.py
```diff
@@ -238,33 +239,43 @@ def statement_display():
 
         # Get statements and assign belief from the meta data
         stmt_iter = []
+        source_counts = {}
         for rel in relations:
+            # Remove statements with only medscan evidence if not logged in
+            if (
+                remove_medscan
+                and "medscan" in rel.data["source_counts"]
+                and set(json.loads(rel.data["source_counts"]).keys()) == {"medscan"}
```
cthoyt, line 248: Since you’re parsing this twice maybe save into a variable

src/indra_cogex/apps/data_display/__init__.py
```diff
@@ -220,6 +220,7 @@ def get_evidence(stmt_hash):
 def statement_display():
     user, roles = resolve_auth(dict(request.args))
     email = user.email if user else ""
+    remove_medscan = user is None
```
cthoyt, line 223: I don’t understand why this is the criteria for removing medscan

src/indra_cogex/apps/data_display/__init__.py
kkaris: I thought it might make the check faster, but thinking twice that's not the case here

src/indra_cogex/apps/data_display/__init__.py
```diff
@@ -238,33 +239,43 @@ def statement_display():
...
             source_counts_per_hash=source_counts,
+            remove_medscan=remove_medscan,
         )
 
         available_sources_dict = {}
         for src_type, sources in sources_dict.items():
-            available_sources_dict[src_type] = [
-                src for src in sources if src in available_sources
-            ]
+            available_sources_dict[src_type] = []
```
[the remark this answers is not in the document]
kkaris: resolved in https://github.com/bgyori/indra_cogex/pull/100/commits/a379bb4600868b4c4fc14ce5f2466e59d91bc0d5

src/indra_cogex/apps/data_display/__init__.py
kkaris: Resolved in https://github.com/bgyori/indra_cogex/pull/100/commits/114e288061792fb6301300d0041926658636e691

bgyori (approved):

PR #102, opened 2022-05-18
GO parent terms in enrichment analysis

kkaris:
This PR implements a helper function that extends gene set mappings of go terms to include child gene sets. The helper function is written to be agnostic of namespace but is currently only used for go - gene set mappings.


kkaris:
Hmm, the test I added is the one that fails. It does pass locally, so it might have to do with the bio ontology.

bgyori (approved):

PR #103, opened 2022-06-01
NIH RePORTER integration

bgyori:
This PR creates an initial implementation for ingesting NIH RePORTER data. It creates a node for each project, and connects it to corresponding publications, clinical trials, and patents. Some issues to follow up on:
- Projects have subprojects but the mappings from each subproject to the parent project are not available so this initial implementation skips any subprojects.
- Publications, trials, and patents are linked to projects via the core project ID (~grant number) but the data is otherwise primarily organized by application ID. There are multiple application IDs corresponding to each project ID and it's not clear which of these the publication belongs to, so this implementation adds all relationships.
- An additional feature would be to create ResearchProject-BioEntity links by doing NER on project abstracts, but this is not part of this PR.


src/indra_cogex/sources/nih_reporter/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+from collections import defaultdict
+import pandas
+import pystow
+from indra_cogex.sources.processor import Processor
+from indra_cogex.representation import Node, Relation
+
+
+# Regular expressions to find files of different types
+fname_regexes = {
+    "project": re.compile(r"RePORTER_PRJ_C_FY(\d+).zip"),
```
cthoyt, line 13: might want to escape the `.` in the file extension since those are technically wildcards

PR #105, opened 2022-06-15
Optimize queries for all relations

bgyori:
This PR optimizes and fixes a bug in the `get_all_relations` function. First `triple_query` is extended to allow different directions for the relation, then the `get_relations` function is extended with a `bidirectional` option. The `get_all_relations` query can then perform a single bidirectional query. This also fixes a bug where self-loops were retrieved twice, now these are only returned once. In addition, `get_all_relations` is refactored to be able to specify both the query node's and the other node's type which speeds up queries significantly.


kkaris:
Nice solution!

PR #109, opened 2022-08-16
DB source exports and ingestion upates

kkaris:
This PR makes significant updates to the indra db processor as well as other processors where the source data structures were either updated or changed in some way.

Merge after #98 

## INDRA DB updates
- The script file `raw_export.py` is added to the `indra_db` submodule in the `sources` module. This script does processing of various dumps from the principal database to output assembled statements from raw statements and reading data.
- The script file `assembly.py` is added to the `indra_db` submodule in the `sources` module. This script uses the output from `raw_export.py` to get source counts and belief scores 
- The processor for `indra_db` is updated to handle the updated source data

## Other Updates
- Bgee: the source data is now loaded with a script instead of manually
- CBioPortal: the source data is now loaded with a script instead of manually
- chembl: duplicated nodes are now skipped
- clinicaltrials: instructions for manually getting the source data was added
- EC: A bug where `'.-'` was trailing identifier strings were fixed
- Sider: previously duplicated nodes and relations are now skipped; More data is retained by adding the bio ontology for name lookup


kkaris:
I think I found the issue leading to missing evidences. While that's investigated, I'm converting the PR to a draft.

bgyori (approved):

PR #110, opened 2022-08-19
Assert column types

kkaris:
This PR implements a check of tsv file headers so as to catch bad data types, e.g. `medscan_only:bool` instead of `medscan_only:boolean`.

### DIFF
--- a/src/indra_cogex/sources/processor.py
+++ b/src/indra_cogex/sources/processor.py
@@ -35,6 +35,35 @@ logger = logging.getLogger(__name__)
 # data stored in /usr/local/var/neo4j/data/databases
 
 
+# See https://neo4j.com/docs/operations-manual/4.4/tools/neo4j-admin/neo4j-admin-import/
+# especially sections 4, 5 and 6
+NEO4J_DATA_TYPES = (
+    "int",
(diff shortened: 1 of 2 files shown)


src/indra_cogex/sources/processor.py
```diff
@@ -231,3 +271,10 @@ def validate_relations(relations):
         except Exception as e:
             logger.info(f"{idx}: {rel} - {e}")
             continue
+
+
+def validate_headers(headers: Iterable[str]) -> None:
+    """Check for data types in the headers"""
+    for header in headers:
+        if ":" in header and header.split(":")[1] not in NEO4J_DATA_TYPES:
```
cthoyt, line 279: does this consider if it ends with a [] for lists?

kkaris: Good point, it does not currently, I'll add that.

kkaris: Resolved in https://github.com/bgyori/indra_cogex/pull/110/commits/d3e95ce3ca85dac861edad7676c55409515168e5

bgyori (approved):

PR #114, opened 2022-09-14
Fix gene list analysis

cthoyt:
This PR does the following:

1. Fixes cypher query for creating HPO cache (capitalization in cypher was wrong)
2. Improves documentation for ORA functions
3. Add prefix counter function (this takes a while to run so it's not included in the front-end)


bgyori (approved):
Looks good, thanks!

PR #115, opened 2022-09-14
Optimize client functions and use CurationCache

kkaris:
This PR does two updates:

- Calls to `indra.sources.indra_db_rest.get_curations` are switched to `CurationCache.get_curations`, with some minor updates
- Several queries to the database are updated to use the new flags added in the latest update of the database content


src/indra_cogex/apps/data_display/__init__.py
```diff
@@ -36,12 +36,27 @@
...
+    ----------
+    curation_list :
+        A list of curations to change the type for the source_hash and
+        pa_hash in.
+    """
+    for curation in curation_list:
+        pa_hash = str(curation["pa_hash"])
+        source_hash = str(curation["source_hash"])
+        curation["pa_hash"] = pa_hash
+        curation["source_hash"] = source_hash
```
bgyori, line 54: Does it make sense to implement this as an in-place operation?

kkaris: Hmm, yeah maybe it's safer to return instead

kkaris: Resolved in https://github.com/bgyori/indra_cogex/pull/115/commits/e3242250267c6a88728b3d6553e7ea86be20d512

bgyori (approved):

PR #119, opened 2022-09-25
Extend enrichment with custom background gene sets

bgyori:
This PR allows calling various enrichment functions with custom background gene sets. It also fixes an issue with handling the default background gene sets (all HGNC IDs) where obsolete IDs were also counted turning out to cause a non-negligible increase in gene counts.

Since the enrichment functions and corresponding utility functions are quite complex at this point, it would be great if we could find a way to unit test these. A blunt option would be to just test them as is, which would run some relatively large queries on the service but would otherwise work. Another option is to pre-produce some cached files and pull them for testing (the disadvantage being that only part of the code will be invoked if caching is used). Yet another option would be to mock the neo4j client, and have functions like `get_go` return some minimal test data structure in the context of each test (here the only disadvantage is that we aren't testing the content of the graph directly). @cthoyt what do you think?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,6 +48,7 @@ install_requires =
     statsmodels
     pandas
     python-dateutil
+    zenodo-client<0.3
 
 include_package_data = True
 python_requires = >=3.7
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,6 @@ install_requires =
     statsmodels
     pandas
     python-dateutil
-    zenodo-client<0.3
 
 include_package_data = True
 python_requires = >=3.7
--- a/src/indra_cogex/client/enrichment/discrete.py
+++ b/src/indra_cogex/client/enrichment/discrete.py
@@ -2,7 +2,7 @@
 
 """A collection of analyses possible on gene lists (of HGNC identifiers)."""
 
-from typing import Iterable, List, Mapping, Optional, Set, Tuple
+from typing import Collection, Iterable, List, Mapping, Optional, Set, Tuple
 
 import numpy as np
 import pandas as pd
@@ -186,7 +186,7 @@ def _do_ora(
 def go_ora(
     client: Neo4jClient,
     gene_ids: Iterable[str],
-    background_gene_ids: Optional[Iterable[str]] = None,
+    background_gene_ids: Optional[Collection[str]] = None,
     **kwargs,
 ) -> pd.DataFrame:
     """Calculate over-representation on all GO terms.
@@ -220,7 +220,7 @@ def go_ora(
 def wikipathways_ora(
     client: Neo4jClient,
     gene_ids: Iterable[str],
-    background_gene_ids: Optional[Iterable[str]] = None,
+    background_gene_ids: Optional[Collection[str]] = None,
     **kwargs,
 ) -> pd.DataFrame:
     """Calculate over-representation on all WikiPathway pathways.
@@ -256,7 +256,7 @@ def wikipathways_ora(
 def reactome_ora(
     client: Neo4jClient,
     gene_ids: Iterable[str],
-    background_gene_ids: Optional[Iterable[str]] = None,
+    background_gene_ids: Optional[Collection[str]] = None,
     **kwargs,
 ) -> pd.DataFrame:
     """Calculate over-representation on all Reactome pathways.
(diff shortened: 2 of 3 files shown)


src/indra_cogex/client/enrichment/discrete.py
```diff
@@ -96,6 +96,7 @@ def count_human_genes(*, client: Neo4jClient) -> int:
     query = """\
         MATCH (n:BioEntity)
         WHERE n.id STARTS WITH 'hgnc'
+        AND NOT n.obsolete = "True"
```
cthoyt, line 99: we should check with @kkaris if we've updated this import entry to be a proper boolean yet or if it's still considered as a string. Then it could be updated to simply

```suggestion
        AND NOT n.obsolete
```

src/indra_cogex/client/enrichment/utils.py
```diff
@@ -71,40 +85,52 @@ def collect_gene_sets(
...
+        # background gene IDs which can change during runtime.
+        if cache_file is not None:
+            with open(cache_file, "wb") as fh:
+                pickle.dump(res, fh)
+            GENE_SET_CACHE[cache_file.as_posix()] = res
+
+    # We now apply filtering to the background gene set if necessary
+    if background_gene_ids:
+        for k, v in res.items():
+            res[k] = {vv for vv in v if vv in background_gene_ids}
```
cthoyt, line 132: `v` is the list of HGNC IDs and `vv` is a single HGNC ID, right?

cthoyt:
I think mocking the client to test `indra_cogex.client.enrichment.collect_gene_sets()` would be a good first step since the code for getting the GO, WikiPathways, Reactome, and HPO all wraps this function just with different cypher queries. Same would go for `indra_cogex.client.enrichment.collect_genes_with_confidence()` which supports the `get_entity_to_regulators()` and `get_entity_to_targets()` functions. The way I would do this is create a subclass of the base neo4j client that overrides `query_tx()` to do results lookup based on the cypher query into a pre-defined dictionary.

We could then add testing for `_do_ora()` that requires no access to neo4j at all and instead uses some pre-defined test gene sets, since this function is generic towards the nested dictionary data structure returned by all of these functions.

One question is do you want testing that checks that all of the caching functionality works properly?

src/indra_cogex/client/enrichment/utils.py
```diff
@@ -206,7 +206,10 @@ def collect_genes_with_confidence(
         curie_to_hgnc_ids = defaultdict(dict)
         max_beliefs: Dict[Tuple[str, str, str], float] = {}
         max_ev_counts: Dict[Tuple[str, str, str], int] = {}
-        for result in client.query_tx(query):
+        query_res = client.query_tx(query)
+        if query_res is None:
```
cthoyt, line 210: if there's an error with execution of the query, then `query_tx` returns None - this commit makes sure that's checked otherwise it's possible to get a more cryptic error when trying to iterate `for result in None`. What kind of error handling should we have here? Should it blow up (with a better error message than just `raise RuntimeError`) or should it potentially also return None so it can be gracefully handled by the caller of `collect_genes_with_confidence`?

src/indra_cogex/client/enrichment/discrete.py
```diff
@@ -186,7 +186,7 @@ def _do_ora(
 def go_ora(
     client: Neo4jClient,
     gene_ids: Iterable[str],
-    background_gene_ids: Optional[Iterable[str]] = None,
+    background_gene_ids: Optional[Collection[str]] = None,
```
cthoyt, line 189: Since the annotation for `Iterable` doesn't imply a size, if you happened to pass a generator object, there will be no implementation of `__len__()` and therefore `len()` will raise a TypeError (or some other kind of exception). Therefore, it's better to use `Collection` here since that does imply that there's a size, with all of the other attributes of an `Iterable`.

src/indra_cogex/client/enrichment/discrete.py
bgyori: `AND NOT n.obsolete` wouldn't work just like that since we have "False" and "True" strings for every node.

cthoyt:
Next steps: mock some of those functions

setup.cfg
```diff
@@ -48,6 +48,7 @@ install_requires =
     statsmodels
     pandas
     python-dateutil
+    zenodo-client<0.3
```
cthoyt, line 51: new release 0.3.1 fixes 3.7 compatibility

bgyori: Thanks!

PR #120, opened 2022-10-07
Use separate sessions

kkaris:
This PR implements separate sessions per query to the database in `query_tx`, just like in [MIRA](https://github.com/indralab/mira/pull/66).

I recreated the transaction error by running the web service locally and submitting large amounts of requests from two separate terminals to the service and verified that the same action did not reproduce the error after the update.

~Left to do~
~The second method in the client, `create_tx`, that also does transactions outside a session context is used for [indexing](https://github.com/bgyori/indra_cogex/blob/main/src/indra_cogex/client/neo4j_client.py#L920-L980) and I have not yet tested that to feel confident enough to replace its current implementation.~

`Neo4jClient.query_tx` _and_ `Neo4jClient.create_tx` _are now implemented using_ `Session.read_transaction` _and_ `Session.write_transaction`, _respectively. This prohibits write transactions when using_ `Neo4jClient.query_tx`.

### DIFF
--- a/src/indra_cogex/client/neo4j_client.py
+++ b/src/indra_cogex/client/neo4j_client.py
@@ -83,15 +83,10 @@ class Neo4jClient:
         query_params :
             Parameters associated with the query.
         """
-        # todo: redo this to use `with ... as tx:` contexts like for query_tx
-        tx = self.get_session().begin_transaction()
-        try:
-            tx.run(query, parameters=query_params)
-            tx.commit()
-        except Exception as e:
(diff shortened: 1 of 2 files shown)


kkaris:
Converted to draft for the time being since we might want to replace the current implementation for `create_tx` as well

bgyori:
@kkaris yes, please add this for create_tx as well. That's not used much in practice but we should do it for consistency.

bgyori (approved):
Great, thanks!

PR #123, opened 2022-10-31
Speed up mesh child terms query

### LINKED ISSUE #122: Slow tests
Some of the tests are very slow and cause actions to take a long time:
```
356.20s call     tests/test_queries.py::test_get_pmids_for_mesh
353.11s call     tests/test_queries.py::test_get_evidence_obj_for_mesh_id
352.95s call     tests/test_queries.py::test_get_stmts_for_mesh_id_w_children
```
Can these tests be simplified to work faster? Or is there some issue with the underlying code that could be optimized to make these queries faster?

kkaris:
This PR resolves #122. The slow part has been identified to be querying for mesh child terms over more than one hop while only using `isa` a relationship type. One query for one of the tests runs for 340 seconds in this case.

For some (still unclear) reason, querying using `-[:isa|partof*1..]->` is about _170x faster_ :exploding_head:  than using just using `-[:isa*1..]->`, regardless if the result is empty or not when using the below query structure:

```
MATCH (c:BioEntity)-[:<rel expr>]->(:BioEntity {id: "mesh:<mesh id>"})
RETURN c.id
```
Below is a table detailing the rough timing in seconds for the above query using different values for `mesh id` and `rel expr`:

| `<rel expr>` | D015002 (has no children) | D007855 (1 direct child and 2 indirect children) |
|---:|---|---|
| `isa` | 0.4 | 0.4 |
| `isa*1..` | ***342*** | ***342*** |
| `isa/partof` | 1.6 | 1.7 |
| `isa/partof*1..` | 2.0 | 2.0 |
| `partof/isa` | 1.6 | 1.7 |
| `partof/isa*1..` | 2.0 | 2.0 |

The changes in this PR makes use of this and changes the relation in the query from `:isa*1..` to `:partof|isa*1..` with some post filtering that takes a negligible amount of time.

### DIFF
--- a/src/indra_cogex/client/queries.py
+++ b/src/indra_cogex/client/queries.py
@@ -1105,14 +1105,36 @@ def _get_mesh_child_terms(
         to the graph.
     """
     meshid_norm = norm_id(*mesh_term)
+    # todo: figure out why [:isa|partof*1..] is ~170x faster than [:isa*1..]
+    #  for the query below
     query = (
         """
-        MATCH (c:BioEntity)-[:isa*1..]->(:BioEntity {id: "%s"})
-        RETURN DISTINCT c.id
+        MATCH p=(c:BioEntity)-[:isa|partof*1..]->(:BioEntity {id: "%s"})
+        RETURN p
     """
         % meshid_norm
     )
-    return set(client.query_tx(query, squeeze=True))
+    paths = client.query_tx(query, squeeze=True)
+    relations_list = [client.neo4j_to_relations(p) for p in paths]
+    mesh_ids = set()
+    for path_result in relations_list:
+        current_set = set()
+        for relation in path_result:
+            # Check if relation is 'isa' and if both nodes are mesh nodes
+            if relation.rel_type == "isa" and \
+                    relation.source_ns.lower() == "mesh" and\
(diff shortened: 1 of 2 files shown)


src/indra_cogex/client/queries.py
```diff
@@ -1105,14 +1105,36 @@ def _get_mesh_child_terms(
...
     )
-    return set(client.query_tx(query, squeeze=True))
+    paths = client.query_tx(query, squeeze=True)
+    relations_list = [client.neo4j_to_relations(p) for p in paths]
+    mesh_ids = set()
+    for path_result in relations_list:
+        current_set = set()
+        for relation in path_result:
+            # Check if relation is 'isa' and if both nodes are mesh nodes
+            if relation.rel_type == "isa" and \
```
kkaris, line 1124: If it's guaranteed that the other node in a `isa` relation from/to a mesh term node always is a mesh term, this filtering is unnecessary.

bgyori: I actually think we could omit all the extra logic here since both `isa` and `partof` are usually legitimate when in comes to finding child terms. In the case of MeSH, `partof` will not appear in practice so we could effectively keep the old code and expect it to work.

bgyori (approved):

PR #125, opened 2022-12-09
New source: InterPro Domains

cthoyt:
This PR imports InterPro entries corresponding to protein domains along with the following relationships for each:

1. `isa` relationships in the InterPro domain hierarchy
2. `has_member` relationships pointing to human proteins (only ones that were mappable via ProtMapper)
3. `associated` relationships to gene ontology terms

<details>
<summary>code to check hierarchical redundancies</summary>

```python
from indra_cogex.sources.interpro import InterproProcessor

proc = InterproProcessor()
proc

c = 0
t = 0
for parent, children in proc.parents.items():
    parent_genes = {gene for gene, _, _ in proc.interpro_to_genes.get(parent, [])}
    if not parent_genes:
        continue
    for child in children:
        child_genes = {gene for gene, _, _ in proc.interpro_to_genes.get(child, [])}
        if not child_genes:
            continue
        
        intersection = parent_genes.intersection(child_genes)
        union = parent_genes.union(child_genes)
        c += 1
        overlap = intersection == child_genes
        t += overlap
        if not overlap:
            print(parent,child, len(child_genes), len(intersection))
        
print(f"{t:,} full intersection ({t/c:.1%}), {c-t:,} missing {(c-t)/c:.1%}")
```

</details>

### DIFF
--- a/docs/source/modules/sources/index.rst
+++ b/docs/source/modules/sources/index.rst
@@ -15,6 +15,7 @@ INDRA CoGEx Sources
    goa
    indra_db
    indra_ontology
+   interpro
    odinson/index
    pathways
    pubmed
--- a/docs/source/modules/sources/interpro.rst
+++ b/docs/source/modules/sources/interpro.rst
@@ -1,0 +1,4 @@
+InterPro Processor (:py:mod:`indra_cogex.sources.interpro`)
+===========================================================
+.. automodule:: indra_cogex.sources.interpro
+    :members:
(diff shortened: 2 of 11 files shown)


bgyori:
Upon further thinking, I realized that the `has_member` relation doesn't have the right semantics (that would be more appropriate for InterPro family relationships but here we are just looking at domains) so I changed the order of the source/target and changed the relation type to `has_domain`.

bgyori (approved):
Nice, thanks!

bgyori:
Another small thing I changed is that technically Node and Relation take prefixes in the INDRA-standard format so e.g., `IP`.

PR #128, opened 2023-01-11
Fix belief bug

kkaris:
This PR handles a bug in the belief score calculation where the wrong variable was used to assign the final score, resulting in only the last calculated belief in a batch of statements being assigned to all statements in the batch.

The bug is fixed and the affected code and its surrounding code is reorganized to make it more clear.

The latest eidos reader version is also added in this PR.


kkaris:
The last commit had to do with networkx, we might want to consider pinning networkx to <3.x for now

PR #131, opened 2023-02-09
Add CellMarker source

cthoyt:
This PR adds the [CellMarker](http://xteam.xbio.top/CellMarker/) resource, which manually curated ~14k marker proteins for tissue/cell pairs. It includes experimental information as well as PubMed literature references for each.

This PR makes the design decision to make the relationships between cells and genes using a new "has_marker" relation, and attribute the tissue type as context on to the relationship since it's not so easy to represent a 3-way relationship without introducing new node types. It also aggregates multiple potential evidences/experimental information/tissues for each cell-gene pair to reduce duplicate relationships.

CC @Siliegia


bgyori (approved):
Similar to the InterPro PR, I had to change namespaces to use the INDRA standard expected by the Node/Relation classes for standardization (e.g., CL->MESH) to work.

PR #134, opened 2023-03-10
Index id property

kkaris:
This PR solves a longstanding mystery of slow lookup when matching against the `id` property of nodes. While it appears that the _internal_ id, accessed via `id(n)` is indexed, the _node property_ `id` was not.

The index building for the id property is added as a separate function to the indexing module and is included when running the import script, `import.sh`.

### DIFF
--- a/src/indra_cogex/indexing/__init__.py
+++ b/src/indra_cogex/indexing/__init__.py
@@ -41,7 +41,7 @@ def index_nodes_on_id(client: Neo4jClient, exist_ok: bool = False):
 
 
 def index_evidence_on_stmt_hash(client: Neo4jClient, exist_ok: bool = False):
-    """Index all Evidence nodes on the statement hash
+    """Index all Evidence nodes on the stmt_hash property
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


bgyori (approved):

kkaris:
Updated the rest api docs examples

src/indra_cogex/client/queries.py
```diff
@@ -870,15 +870,15 @@ def get_evidences_for_stmt_hashes(
 
 @autoclient()
 def get_stmts_for_paper(
-    term: Tuple[str, str], *, client: Neo4jClient, **kwargs
+    pmid_term: Tuple[str, str], *, client: Neo4jClient, **kwargs
```
cthoyt, line 873: does this name make sense? you can give PMC, TRID, or DOI here too

kkaris: That update was needed to fix the api docs

kkaris: The dict that populates the api doc examples is keyed by the argument names

kkaris: But we can change `pmid_term` to `paper_id` or something similar

kkaris: Resolved in https://github.com/bgyori/indra_cogex/pull/134/commits/21ebf82b7d9a885e45f591581e5230719728b15a

PR #136, opened 2023-05-18
Dump updates

kkaris:
This PR collects minor updates to two processors:

1. The pre-processor process for indra db - the assembly and subsequent refinement and belief calculation - has been updated with new time estimates and counts to keep up with the growing size of the content.
2. The pubmed processor has been updated to not rely on hard coded indices from the download site.


bgyori (approved):
Nice, thanks!

bgyori (approved):
Nice, thanks!

PR #138, opened 2023-05-26
Processor & related updates

kkaris:
In this PR:
- The HPOA processors are updated to handle new format of the source data
- Node de-duplication and special character escaping is done for Nodes in the NIH reporter
- Data types are added to the CCLE, ChEMBL and NIH reporter processor output
- Labels are added to the discovery.indra.bio landing page for the two new edge labels (Cell Markers and InterPro Domains)
- Example input in the API docs is added for `get_stmts_for_pubmeds`

The data processed with this code is already in the database, but the service running at discovery.indra.bio needs to be restarted with the new code.

### DIFF
--- a/src/indra_cogex/apps/constants.py
+++ b/src/indra_cogex/apps/constants.py
@@ -30,6 +30,8 @@ edge_labels = {
     "has_publication": "Project Publications",
     "has_clinical_trial": "Project Trials",
     "has_patent": "Project Patents",
+    "has_marker": "Cell Markers",
+    "has_domain": "InterPro Domains",
 }
 
 INDRA_COGEX_WEB_LOCAL = (get_config("INDRA_COGEX_WEB_LOCAL") or "").lower() in {
--- a/src/indra_cogex/apps/constants.py
+++ b/src/indra_cogex/apps/constants.py
@@ -31,7 +31,7 @@ edge_labels = {
     "has_clinical_trial": "Project Trials",
     "has_patent": "Project Patents",
     "has_marker": "Cell Markers",
-    "has_domain": "InterPro Domains",
+    "has_domain": "Protein Domains",
 }
 
 INDRA_COGEX_WEB_LOCAL = (get_config("INDRA_COGEX_WEB_LOCAL") or "").lower() in {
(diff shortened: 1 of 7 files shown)


src/indra_cogex/apps/constants.py
```diff
@@ -30,6 +30,8 @@
     "has_publication": "Project Publications",
     "has_clinical_trial": "Project Trials",
     "has_patent": "Project Patents",
+    "has_marker": "Cell Markers",
+    "has_domain": "InterPro Domains",
```
cthoyt, line 34:
```suggestion
    "has_domain": "Protein Domains",
```

src/indra_cogex/sources/chembl/__init__.py
```diff
@@ -91,7 +91,7 @@ def get_relations(self) -> Iterable[Relation]:
                 "has_indication",
                 {
                     "source": self.name,
-                    "max_phase:int": max_phase,
+                    "max_phase:float": max_phase,
```
cthoyt, line 94: why float? to handle nans? typically this is 0-4

src/indra_cogex/sources/nih_reporter/__init__.py
```diff
@@ -241,6 +259,17 @@ def download_files(
             )
 
 
+def clean_text(text: Any) -> Any:
```
cthoyt, line 262: this function must be rewritten several places throughout the project if you want to try and wrangle them

src/indra_cogex/sources/chembl/__init__.py
kkaris: Because some phases are =0.5

kkaris: See https://chembl.gitbook.io/chembl-interface-documentation/frequently-asked-questions/drug-and-compound-questions#what-is-max-phase

src/indra_cogex/sources/nih_reporter/__init__.py
kkaris: This is the only place so far (to my knowledge) where this is needed, so maybe we can keep it as is for now?

cthoyt: Yes fine for now :)

bgyori (approved):
Looks great, thanks!

PR #145, opened 2023-10-19
add odinson and gilda grounding for spine

svempati21:


bgyori:
I just noticed - this work doesn't really belong in this repository. It should be in the https://github.com/gyorilab/indra_spine repository since that's where this line of work has been developed.

PR #147, opened 2023-12-11
Add DisGeNet processor

cthoyt:
This branch is 2 years old, but we're back ;) 

This PR adds the following processors for manually curated information in DisGeNet:

- [x] disease-gene associations
- [x] disease-variant (i.e., SNP) associations
- [x] gene-variant associations 

Because DisGeNet uses UMLS to refer to genes, it externalizes the UMLS mapping tooling that was implemented for SIDER so it can be more easily reused

### DIFF
--- a/src/indra_cogex/sources/disgenet/__init__.py
+++ b/src/indra_cogex/sources/disgenet/__init__.py
@@ -1,0 +1,222 @@
+# -*- coding: utf-8 -*-
+
+"""This script gets DisGeNet, converts to BEL, exports in a few formats."""
+
+import logging
+import pickle
+
+import click
+import pandas as pd
+import pystow
+from indra.databases import hgnc_client
+
+from indra_cogex.representation import Node, Relation
+from indra_cogex.sources import Processor
+from indra_cogex.sources.utils import UmlsMapper
+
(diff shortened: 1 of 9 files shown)


src/indra_cogex/sources/disgenet/__init__.py
```diff
@@ -0,0 +1,186 @@
...
+            yield Node.standardized(
+                db_ns=disease_class, db_id=identifier, name=name, labels=["BioEntity"]
+            )
+        for snp_id in self.variant_df["snpId"].unique():
+            yield Node.standardized(
+                db_ns="DBSNP", db_id=snp_id, name=snp_id, labels=["BioEntity"]
+            )
+
+    def get_relations(self):  # noqa:D102
+        # Use diseaseName instead of diseasePrefix as diseasePrefix does not
```
cthoyt, line 134: the point of the UMLS mapper was to add standardized fields `disease_prefix`, `disease_id`, and `disease_name`, which aren't the same as the camel-cased ones

a trick for when stuff isn't what you expect is to print(df.head()) to get a look at some sample data

bgyori (approved):

PR #155, opened 2024-01-31
Replacing sleep with poll

jim-sheldon:
Get the HTTP port from Neo4J config file, `curl` endpoint on `localhost` until ready.


bgyori (approved):
This is great, thanks!

PR #149, opened 2024-01-10
Dumping updates and data validation

kkaris:
This PR adds the latest iteration of updates related to running the full dump after dependencies have been updated. It also adds data validation to the processors.

Dump and content updates:
- PublicationType tags are added to Publication nodes, following [updates in INDRA](https://github.com/sorgerlab/indra/pull/1428)
- A boolean is added to `Publication`, `Evidence` nodes and `indra_rel` relationships which is true if Publication/Evidence is retracted or if a statement has at least one Evidence from a retracted source
- Add a helper function to get the proper Neo4j boolean values from a Python boolean or condition and use it in all places that generate boolean metadata
- Put locations for processor files in their own files in `indra_db` and `pubmed` sub modules to avoid circular imports
- A typo for a key previously lead to indra evidence jsons in Evience nodes not being validated, the typo is now fixed (see `assert_valid_node()`)
- Beliefs are set in the statement json of `indra_rel` relations. Previously, this was only done for the belief _property_ of the `indra_rel` relations
- Various improvements to logging and comments

Data validation:
- The data validation checks that a value and its type are as expected for a given Neo4j data type as specified in the ingestion tsv file header. See [the documentation](https://neo4j.com/docs/operations-manual/4.4/tools/neo4j-admin/neo4j-admin-import/#import-tool-header-format-properties) for more info on Neo4j data types.
- The data validation is done at dump time and is added in `assert_valid_node()`, which is called from both `validate_nodes()` and `validate_relations()`.
- Tests are added for the data validator

todo:
- [x] Process and dump nodes edges from `indra_db`
- [x] Fix missing data for Publication nodes
- [x] Rectify inconsistencies in return type (source counts vs evidence counts) in `curator_blueprint.py`
- [x] Fix instances where duplicate nodes are generated (check import.report for hints). Duplication can happen either at the processor level (e.g. a node appears multiple times) or at the assembly level (e.g. the same node is produced by different processors).
- [x] Check unicode issue (see #141 and #142).

PR #156, opened 2024-02-05
Remove scipy dependency constraint

bgyori:
The issue https://github.com/scipy/scipy/issues/16079 has been resolved so we can remove the version constraint.


kkaris:
The two failed tests are fixed in #149

Issue #161, opened 2024-05-06
Discrete GSEA issues


bgyori:
When using https://discovery.indra.bio/gene/discrete
- Often get timeout (504 error) with the default gene list, if including INDRA path-based analysis
- The result lists cannot be downloaded in any form
- There is no way to click through to see the statements behind INDRA upstream/downstream entities - this could be added by linking to subnetwork curation

kkaris:
> Often get timeout (504 error) with the default gene list, if including INDRA path-based analysis

Quick workaround: Set minimum evidence needed to 2 or higher and the analysis should finish within the 60 s timeout. Setting the belief cutoff at 0.6 seems to work as well.

> The result lists cannot be downloaded in any form

This can be turned on by setting `INDRA_COGEX_WEB_LOCAL=TRUE` in the environment. See https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/apps/constants.py#L47-L50

PR #162, opened 2024-05-09
Fixes related to #160, #161

### LINKED ISSUE #160: Subnetwork curation issues
At https://discovery.indra.bio/curate/subnetwork:
- curation doesn't work even after logging in
- CURIEs have to use the INDRA standard (e.g., HGNC:6192, CHEBI:CHEBI:71200) which would normally be fine but it's a bit inconsistent since other parts of the website assume Bioregistry curies like chebi:71200
- When no statements are found, we get a message saying "There are no statements for this publication." which is not quite right since we aren't looking at publications

kkaris:
This PR fixes some of the issues mentioned in #160:

- The option to pass a message when no statements are found is added as well as a more generic default message
- `process_identifier()` is done on the curies passed from the subnetwork form before passing them on to allow for some flexibility in how the curies are formatted
- `INDRA_COGEX_WEB_LOCAL` is passed as a build ARG in the Dockerfile to allow for downloads to be turned on

Curations are working again since #160 was posted.

Resolves #160.

Work on linking out to the statements when using INDRA up/downstream GSEA mentioned in #161 is ongoing work.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,6 +3,7 @@ FROM labsyspharm/indra
 ARG INDRA_NEO4J_URL
 ARG INDRA_NEO4J_USER
 ARG INDRA_NEO4J_PASSWORD
+ARG INDRA_COGEX_WEB_LOCAL
 
 RUN pip3 install git+https://github.com/bgyori/indra_cogex.git#egg=indra_cogex[web,gunicorn,gsea] && \
     python -m indra.ontology.bio build && \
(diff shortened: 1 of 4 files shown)


kkaris:
I found the pull request where Charlie implemented the download: https://github.com/gyorilab/indra_cogex/pull/69

He says somethong about this creating a file on the host machine, which would mean that there are files generated in the EC2 instance, I need to look into this before we merge this PR.

PR #168, opened 2024-06-03
Pubmed rels debug

kkaris:
This PR splits the `PubmedProcessor` into two separate processors: `PublicationProcessor` and `JournalProcessor`. This is done in order to not have a conflict in the (previously) common edges file.

Other updates:
- Resolve some data conflict in journal nodes from Pubmed and the wikidata `JournalPublisherProcessor`
- Various changes to catch NaN's in `JournalPublisherProcessor`
- Add option to validate data in `Node` creation, turned off by default, turned on for Node assembly
- Replace `pyobo` with `bio_ontology` where possible
- For cases where `pyobo` is used, specify version to avoid downloads of resource file when service is running
- Add cli for pickle caches

**Todo:**

- [x] Update Dockerfile to copy over pyobo resource files and to rebuild caches
- [ ] Rebuild and re-deploy service after merge

### DIFF
--- a/src/indra_cogex/apps/constants.py
+++ b/src/indra_cogex/apps/constants.py
@@ -43,6 +43,7 @@ edge_labels = {
     "variant_gene_association": "Variant Gene Associations",
     "variant_phenotype_association": "Variant Phenotype Associations",
     "has_activity": "Enzyme Annotations",
+    "published_by": "Journal-Publisher Associations",
 }
 
 INDRA_COGEX_WEB_LOCAL = (get_config("INDRA_COGEX_WEB_LOCAL") or "").lower() in {
--- a/src/indra_cogex/apps/constants.py
+++ b/src/indra_cogex/apps/constants.py
@@ -109,3 +109,8 @@ else:
         "and CLARE_PUSHER_CLUSTER."
     )
     pusher_app = None
+
+PYOBO_RESOURCE_FILE_VERSIONS = {
+    "mgi": "6.23",
+    "rgd": "2024-05-31",
+}
--- a/src/indra_cogex/apps/templates/home.html
+++ b/src/indra_cogex/apps/templates/home.html
@@ -42,7 +42,13 @@
     <div class="col" style="margin-bottom: 1em">
         {% set count, suffix = format_number(counter[key]) %}
         <span class="count-{{ key }}" data-to="{{ count }}" data-time="1000"></span>{{ suffix }}<br/>
-        <small>{{ label|replace(" ", "&nbsp;")|safe }}</small>
+        <small>
+            {% if label %}
+                {{ label|replace(" ", "&nbsp;")|safe }}
+            {% else %}
+                {{ key }}
+            {% endif %}
+        </small>
     </div>
 {% endmacro %}
 
--- a/src/indra_cogex/client/enrichment/mla.py
+++ b/src/indra_cogex/client/enrichment/mla.py
@@ -11,6 +11,7 @@ import indra.statements
 import pandas as pd
 import pyobo
 from indra.databases.hgnc_client import hgnc_to_enzymes
+from indra.ontology.bio import bio_ontology
 from indra.statements import stmts_from_json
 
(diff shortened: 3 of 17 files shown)


kkaris:
There's a circular import breaking the tests, I'll fix it

src/indra_cogex/apps/constants.py
```diff
@@ -108,3 +109,8 @@
         "and CLARE_PUSHER_CLUSTER."
     )
     pusher_app = None
+
+PYOBO_RESOURCE_FILE_VERSIONS = {
+    "mgi": "6.23",
+    "rgd": "2024-05-31",
+}
```
bgyori, line 116: Depending on exact usage, these could be replaced by what is already available via `indra.databases`

src/indra_cogex/sources/pubmed/__init__.py
```diff
@@ -119,67 +213,30 @@ def _yield_journal_nodes(self) -> Iterable[Node]:
...
-
-    def _yield_pmid_journal_relations(self) -> Iterable[List[Relation]]:
+    def _get_relations(self) -> Iterable[List[Relation]]:
         # Yield batches of relations
         with gzip.open(self.pmid_nlm_path, "rt") as fh:
             reader = csv.reader(fh)
             # Skip header
             next(reader)
-            # The file has more than 35000000 lines - a batch size of 1M is
+            # The file has more than 3500000 lines - a batch size of 1M is
```
bgyori, line 239: This could be a typo?

kkaris: Resolved in 39bb0662540c859c1216207d3803c6a1c6627616

src/indra_cogex/apps/constants.py
kkaris: The usage is to [look up identifiers from names/symbols](https://github.com/gyorilab/indra_cogex/blob/pubmed-rels-debug/src/indra_cogex/client/enrichment/continuous.py#L183), but not translate them to human equivalents. So I think we'll keep the implementation with pinning the versions and just make sure the resource files exist as part of the docker build.

src/indra_cogex/client/enrichment/mla.py
```diff
@@ -78,12 +78,18 @@ def get_metabolomics_sets(
...
     """
     )
-    for ec_curie, chebi_curies in client.query_tx(query):
+    for ec_curie, ec_name, chebi_curies in client.query_tx(query):
         ec_code = ec_curie.split(":", 1)[1]
-        rv[ec_code, pyobo.get_name("ec", ec_code)].update(
+        # There are a few cases where the name is not in the database, try to get it
+        # from the bio_ontology in those cases
+        if not ec_name:
+            name = bio_ontology.get_name("ECCODE", ec_code)
```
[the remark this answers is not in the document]
kkaris: Fixed in e4a5e9978f5464ce7e13f9a7091387b1498e5f6e

kkaris:
I tried to run the continuous analysis with some made up rat data (just listing some rat gene names and associating random floats with each name), but the GSEA algorithm filtered it all out and errored so it's still untested. The file copying at docker build time and loading of the name to id map worked out though.

src/indra_cogex/apps/templates/home.html
```diff
@@ -42,7 +42,13 @@ <h2><span class="count-{{ key }}" data-to="{{ count }}" data-time="1000"></span>
     <div class="col" style="margin-bottom: 1em">
         {% set count, suffix = format_number(counter[key]) %}
         <span class="count-{{ key }}" data-to="{{ count }}" data-time="1000"></span>{{ suffix }}<br/>
-        <small>{{ label|replace(" ", "&nbsp;")|safe }}</small>
+        <small>
```
kkaris, line 45: I added this so it's really easy to see which key is missing a corresponding label, but if we would rather have blank space when something is missing, we can revert this change.

bgyori (approved):

PR #172, opened 2024-06-07
Upgrade to flask_jwt_extended>3

kkaris:
Update usage of `flask_jwt_extended` decorator to be compatible with flask_jwt_extended>3, which is a requirement for using Flask>=3.

See https://github.com/sorgerlab/indra/issues/1448.


kkaris:
The failing test is because I restored setup.cfg to again refer to the non-dev branches of indra and ui util.

Issue #83, opened 2022-03-23
Add gif/spinner while backend is loading g/msea results


kkaris:
When the g/msea pages are crunching the numbers:

1. show a spinner of sorts to notify the user that "we're working on it"
2. disable the submit button, I think the behavior now is that the page crashes if it's clicked another time while work is ongoing

PR #183, opened 2024-09-27
Add gif/spinner while backend is loading g/msea results/ Issue #83

prasham21:
This PR adds a spinner next to the submit button in the form. The spinner is displayed when the form is submitted to indicate loading status, and the submit button is disabled to prevent multiple submissions during the loading period. This improves user experience by providing visual feedback during form processing.

### DIFF
--- a/src/indra_cogex/apps/templates/base_form.html
+++ b/src/indra_cogex/apps/templates/base_form.html
@@ -34,3 +34,4 @@
     });
 </script>
 {% endblock %}
+
--- a/src/indra_cogex/apps/templates/base_form.html
+++ b/src/indra_cogex/apps/templates/base_form.html
@@ -1,13 +1,36 @@
 {% extends "base.html" %}
-
 {% import "bootstrap4/form.html" as wtf %}
-
 {% block container %}
     <div class="card card-body bg-light">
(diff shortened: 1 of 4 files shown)


src/indra_cogex/apps/templates/base_form.html
```diff
@@ -1,13 +1,37 @@
...
 {% block container %}
     <div class="card card-body bg-light">
         <h1 class="display-4">{% block header %}{% endblock %}</h1>
         <p class="lead">
             {% block lead %}{% endblock %}
         </p>
-        {{ wtf.render_form(form) }}
+        <form id="main-form" method="POST">
+            {{ wtf.render_form(form) }}
+        </form>
```
kkaris, line 11: I'm not sure you need to add this, the form tag is autogenerated by `wtf.render_form` and you can always select the form element using [class name](https://www.w3schools.com/jsref/met_document_getelementsbyclassname.asp) or by [tag name](https://www.w3schools.com/jsref/met_document_getelementsbytagname.asp).
Another option is to look into if you can add the `id` as an extra parameter to `wtf.render_form`.

prasham21: Fixed in 73558979

PR #186, opened 2024-10-28
Endpoint Queries

haohangyan:
This PR adds the REST API related to getting statements from graph databases based on various constraints
There are four endpoints added:
get_stmts_for_agent_type: Query Retrieves statements on agent type (subject/subject) as well as the agent name in text
get_stmts_for_source: Retrieves statements based on the source type (e.g "reach")
get_stmts_for_rel_type: Retrieves statements based on the statement type(e.g. "Phosphorylation")
get_statements_mix: A comprehensive endpoint that allows combined filtering by agent name, agent type, source type, and statement type.

## Todo

In the `get_statements_mix` function:
- [x] Rename the function to simply `get_statements`
- [x] `agent_name` should be `agent` and be either a tuple, which assumes (namespace, id) is provided, or a string, which assumes a name is provided.
- [x] Add `other_agent` and `other_role` as optional parameters for second agent in query. 
- [x] Add MeSH filter, this is missing currently.
- [x] Add paper filter, this is also missing currently.


src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,246 @@ def get_stmts_for_stmt_hashes(
...
+        The source to query for (e.g., "reach").
+    agent_name : Optional[str], default: None
+        The name of the agent to filter by (e.g., "EGFR").
+    agent_role : Optional[str], default: None
+        The role of the agent in the interaction, either "subject" or "object".
+    limit : Optional[int], default: 10
+
+    client : Neo4jClient
+
+    evidence_limit : Optional[int], default: None
```
kkaris, line 1173: Add descriptions for these two parameters for completeness (there are more examples of this, but I'm just gonna comment on this one):
```suggestion
    client : Neo4jClient
        text
    evidence_limit : Optional[int], default: None
        text
```

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,246 @@ def get_stmts_for_stmt_hashes(
...
+        "limit": limit
+    }
+
+    logger.info(f"Running query with constraints: rel_type={rel_type}, "
+                f"source={stmt_source}, agent_name={agent_name}, "
+                f"agent_role={agent_role}, limit={limit}")
+    rels = client.query_relations(query, **params)
+    stmts = indra_stmts_from_relations(rels, deduplicate=True)
+
+    if evidence_limit:
```
kkaris, line 1223: There is already one evidence in the `indra_rel` relations, so it only makes sense to fetch more evidence if the limit is `>1`.
```suggestion
    if evidence_limit and evidence_limit > 1:
```

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,246 @@ def get_stmts_for_stmt_hashes(
...
+        MATCH p = (a:BioEntity {name: $agent_name})-[r:indra_rel]->(b:BioEntity)
+        RETURN p LIMIT $limit
+        """
+    elif agent_role.lower() == "object":
+        query = """
+        MATCH p = (a:BioEntity)-[r:indra_rel]->(b:BioEntity {name: $agent_name})
+        RETURN p LIMIT $limit
+        """
+    else:
+        raise ValueError("agent_role must be 'subject' or 'object'")
```
kkaris, line 1273: I think it should be possible to allow the agent to be either of `subject` or `object` as well. That query would either be a non-directed query: `()-[]-()` or a union of two directed queries `()-[]->()` and `()<-[]-()` and then ensuring there are no duplicates in the results.

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,246 @@ def get_stmts_for_stmt_hashes(
...
+        limit: Optional[int] = 10,
+        *,
+        client: Neo4jClient,
+        evidence_limit: Optional[int] = None
+) -> List[Statement]:
+    """Return the statements for the given source.
+
+    Parameters
+    ----------
+    stmt_source : str
```
kkaris, line 1303: We should think about allow a list of sources rather than just one.

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,246 @@ def get_stmts_for_stmt_hashes(
...
+        limit: Optional[int] = 10,
+        *,
+        client: Neo4jClient,
+        evidence_limit: Optional[int] = None
+) -> List[Statement]:
+    """Return the statements for the given relationship type.
+
+    Parameters
+    ----------
+    rel_type : str
```
kkaris, line 1351: Same here, multiple relationship types should be allowed

kkaris (changes requested):
I think this looks good overall, I only have minor changes that I think you should look at. Otherwise, I have some suggestions that we should discuss if they make sense or not that we don't necessarily have to implement.

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,274 @@ def get_stmts_for_stmt_hashes(
...
+    elif agent_role.lower() == "object":
+        query = """
+        MATCH p = (a:BioEntity)-[r:indra_rel]->(b:BioEntity {name: $agent_name})
+        RETURN p LIMIT $limit
+        """
+    else:
+        query = """
+        MATCH p = (a:BioEntity {name: $agent_name})-[r:indra_rel]-(b:BioEntity)
+        RETURN p LIMIT $limit
+        """
```
kkaris, line 1289: This could be simplified or replaced with use of [`triple_parameter_query`](https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/representation.py#L285-L371) in `indra_cogex.representation` and using the `relation_direction` parameter of `triple_parameter_query`.

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,274 @@ def get_stmts_for_stmt_hashes(
...
+        stmt.get_hash(): rel.data["evidence_count"]
+        for rel, stmt in zip(rels, stmts)
+    }
+
+    return stmts, evidence_counts
+
+@autoclient()
+def get_stmts_for_agent_type(
+    agent_name: str,
+    agent_role: str,
```
kkaris, line 1247: If it's allowed to not specify a direction, this parameter should be optional.
```suggestion
    agent_role: Optional[str] = None,
```

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1144,274 @@ def get_stmts_for_stmt_hashes(
...
+    evidence_limit: Optional[int] = None
+) -> List[Statement]:
+    """Return the statements for a given agent based on its role
+
+    Parameters
+    ----------
+    agent_name : str
+        The name of the agent (e.g.,"MEK").
+    agent_role : str
+        The role of the agent in the interaction, either "subject" or "object".
```
kkaris, line 1260: Add that `agent_role` is optional.

kkaris (changes requested):
Just minor changes requested, otherwise it's done.

src/indra_cogex/client/queries.py
haohangyan: I feel like using triple_parameter_query requires more parameter names and codes to construct the queries.

kkaris:
Updated description with todo list

src/indra_cogex/client/queries.py
```diff
@@ -1140,6 +1127,345 @@ def get_stmts_for_stmt_hashes(
...
+        rel_types: Optional[Union[str, List[str]]] = None,
+        stmt_sources: Optional[Union[str, List[str]]] = None,
+        agent_role: Optional[str] = None,
+        other_agent: Optional[Union[str, Tuple[str, str]]] = None,
+        other_role: Optional[str] = None,
+        paper_term: Optional[Tuple[str, str]] = None,
+        mesh_term: Optional[Tuple[str, str]] = None,
+        include_child_terms: Optional[bool] = True,
+        limit: Optional[int] = 10,
+        client: Neo4jClient,
```
kkaris, line 1143: I think the ordering here works because of the autoclient decorator, but we should still move the `client` keyword argument to before the optional ones.

PR #191, opened 2024-12-16
Add missing queries

prasham21:
Added functions for all missing relationships in the queries module. For each relationship, implemented:
get_targets() function to get target nodes
get_sources() function to get source nodes
has_relation() function to check if relationship exists

Also added corresponding tests for all new functions to verify they work correctly with the database. 

Also, this branch contains: 
Reorganize API endpoints into functional categories
Group endpoints based on functionality (search, validation, evidence, etc.)
Add category-specific namespaces with descriptive documentation
Maintain existing endpoint functionality.


src/indra_cogex/apps/rest_api.py
kkaris, line 1: `CATEGORY_DESCRIPTIONS` and all the namespaces created in this file should be moved to `queries_web.py`. The import structure of these file should be that you define `Namespace` objects in the file where you create the endpoints, in this case `queries_web/__init__.py`, then in `rest_api.py` you import those `Namespace` objects and add them to the api. Now it's the other way around.

src/indra_cogex/client/queries.py
```diff
@@ -1740,6 +1794,1447 @@ def _filter_out_medscan_evidence(
...
+    MATCH (c:BioEntity {id: $cell_id})-[r:has_marker]->(m:BioEntity {id: $marker_id})
+    RETURN COUNT(r) > 0 as exists
+    """
+    return client.query_tx(query, cell_id=cell_id, marker_id=marker_id, squeeze=True)[0]
+
+
+# HPOA Functions
+@autoclient()
+def get_phenotypes_for_disease(
+    disease: Tuple[str, str], *, client: Neo4jClient
```
kkaris, line 1882: See if you can find an example for `disease` that returns a result for this function in the API docs, right now the result is empty.

src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -106,14 +279,39 @@ def get_example_data():
...
+    "journal": fields.List(fields.String, example=["nlm", "0000201"]),
+    # Disgenet
+    "variant": fields.List(fields.String, example=[["dbsnp", "rs9994441"], ["dbsnp", "rs13015548"]]),
+    # Wikidata
+    "publisher": fields.List(fields.String, example=["isni", "0000000031304729"]),
+    # NIH Reporter
+    "project": fields.List(fields.String, example=["nihreporter.project", "2106659"]),
+    "patent": fields.List(fields.String, example=["google.patent", "US5939275"]),
+    # HPOA
+    "phenotype": fields.List(fields.String, example=[["hp", "0003138"], ["mesh", "D001827"]]),
```
kkaris, line 304: I think this is wrong, all functions using `phenotype` as parameter in `queries.py` are expecting a singe curie, i.e. `["mesh", "D001827"]` or similar, not a list of curies like this example.
```suggestion
    "phenotype": fields.List(fields.String, example=["hp", "0003138"]),
```
Please also try also to find an example that yields a result for most of the queries using this as an example in the apidocs.

src/indra_cogex/apps/rest_api.py
prasham21: I earlier did that because of a circular import issue. 
But I have fixed it now. 

Fixed in de7bcbd7

src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -106,14 +279,39 @@ def get_example_data():
...
     "species": fields.String(example="human"),
     "permutations": fields.Integer(example=100),
     "source": fields.String(example="go"),
     "indra_path_analysis": fields.Boolean(example=False),
+    # Cell marker
+    "cell_type": fields.List(fields.String, example=["cl", "0000020"]),
+    "marker": fields.List(fields.String, example=["hgnc", "11337"]),
+    # Pubmed
+    "publication": fields.List(fields.String, example=["pubmed", "14334679"]),
+    "journal": fields.List(fields.String, example=["nlm", "0000201"]),
```
kkaris, line 329: Try to find an example that returns a value, this example returns an empty result for me.

src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -106,14 +279,39 @@ def get_example_data():
...
+    # Disgenet
+    "variant": fields.List(fields.String, example=[["dbsnp", "rs9994441"], ["dbsnp", "rs13015548"]]),
```
kkaris, line 331: Same problem here as with `phenotype` below, the expected value is a curie, not a list of curies.
```suggestion
    "variant": fields.List(fields.String, example=["dbsnp", "rs9994441"]),
```

src/indra_cogex/client/queries.py
```diff
@@ -1740,6 +1794,1447 @@ def _filter_out_medscan_evidence(
...
+    return client.get_sources(
+        domain,
+        relation="has_domain",
+        source_type="BioEntity",
+        target_type="BioEntity"
+    )
+
+
+@autoclient()
+def has_domain(
```
kkaris, line 2662: I think we can get a better name than `has_domain`, maybe `gene_has_domain`?

src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -106,14 +279,39 @@ def get_example_data():
...
+    "patent": fields.List(fields.String, example=["google.patent", "US5939275"]),
+    # HPOA
+    "phenotype": fields.List(fields.String, example=[["hp", "0003138"], ["mesh", "D001827"]]),
+    # For InterPro
+    "domain": fields.List(fields.String, example=["interpro", "IPR006047"]),
+    # For DepMap codependency
+    "gene1": fields.List(fields.String, example=["hgnc", "1234"]),
+    "gene2": fields.List(fields.String, example=["hgnc", "5678"]),
+    # For ChEMBL
+    "molecule": fields.List(fields.String, examples=[["chebi", "10001"], ["chembl.compound", "CHEMBL25"]]),
```
kkaris, line 345: I think this is a typo, and just like above, we need to use only one of the curies.
```suggestion
    "molecule": fields.List(fields.String, example=["chebi", "10001"]),
```

src/indra_cogex/client/queries.py
```diff
@@ -1740,6 +1794,1447 @@ def _filter_out_medscan_evidence(
...
+                                                              0].lower() == "mesh" else f"{indication[0]}:{indication[1]}".lower()
+    query = """
+    MATCH (m:BioEntity)-[r:has_indication]->(i:BioEntity {id: $indication_id})
+    RETURN m
+    """
+    return client.query_nodes(query, indication_id=indication_id)
+
+
+@autoclient()
+def has_indication(
```
kkaris, line 2827: We could use a better name here too, maybe `molecule_has_indidcation`.

src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -75,11 +248,11 @@ def get_example_data():
     "agent": fields.String(example="MEK"),
     "other_agent": fields.String(example="ERK"),
     "agent_role": fields.String(example="Subject"),
-    "other_role" : fields.String(example="Object"),
+    "other_role": fields.String(example="Object"),
     "stmt_source": fields.String(example="reach"),
     "stmt_sources": fields.List(fields.String, example=["reach", "sparser"]),
     "include_db_evidence": fields.Boolean(example=True),
-    "cell_line": fields.List(fields.String, example=["CCLE", "BT20_BREAST"]),
+    "cell_line": fields.List(fields.String, example=["CCLE", "BT20_BREAST", "HEL_HAEMATOPOIETIC_AND_LYMPHOID_TISSUE"]),
```
kkaris, line 289: If you want to change the example you need to replace the identifier, not add to the list.
```suggestion
    "cell_line": fields.List(fields.String, example=["CCLE", "HEL_HAEMATOPOIETIC_AND_LYMPHOID_TISSUE"]),
```
So keep either of `"BT20_BREAST"` or `"HEL_HAEMATOPOIETIC_AND_LYMPHOID_TISSUE"`, but not both.

src/indra_cogex/client/queries.py
prasham21: FIxed in 4640b451

src/indra_cogex/apps/queries_web/__init__.py
prasham21: Fixed in 4640b451

src/indra_cogex/client/queries.py
prasham21: Fixed in 4640b451

src/indra_cogex/apps/queries_web/__init__.py
prasham21: Fixed in 4640b451

src/indra_cogex/client/queries.py
prasham21: Fixed in 4640b451

src/indra_cogex/client/queries.py
```diff
@@ -1740,6 +1794,1447 @@ def _filter_out_medscan_evidence(
...
+    gene_id = f"{gene[0]}:{gene[1]}".lower()
+    query = """
+    MATCH (g1:BioEntity {id: $gene_id})-[r:codependent_with]->(g2:BioEntity)
+    RETURN g2
+    """
+    return client.query_nodes(query, gene_id=gene_id)
+
+
+@autoclient()
+def has_codependency(
```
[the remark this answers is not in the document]
prasham21: Fixed in 4640b451

src/indra_cogex/apps/queries_web/__init__.py
prasham21: Fixed in 4640b451

src/indra_cogex/apps/queries_web/__init__.py
prasham21: I think these examples do return values.  I ran a test for these examples. 
marker : ["hgnc", "11337"] returns Cell Type: ["cl", "0000020"] 

publication : ["pubmed", "14334679"] returns  Journal result: [(:Journal { id:'NLM:0000201', issn_l:'0567-8064', name:'Acta radiologica: therapy, physics, biology', abbr_title:'Acta Radiol Ther Phys Biol' })]. 

journal : ["nlm", "0000201"] returns several publications from journal articles.

kkaris: I meant to only comment on the `nlm` example. I think the minimum number of lines shown is 4, so I selected line 295 for the nlm id, but line 292-294 are also shown.

I checked again, and this example does indeed return a result for `/api/get_publications_for_journal`, but it does not for `/api/get_publisher_for_journal`, if you can find an example NLM ID that works for both these endpoints, that would be great, but don't spend too much time on it.

PR #193, opened 2025-01-06
Add multi-example support for API parameters to handle empty results

prasham21:
Adds MultiExample class and updates examples_dict to support multiple example values per parameter. Modifies QueryResource to try alternative examples when initial query returns empty results.

### DIFF
--- a/src/indra_cogex/apps/queries_web/__init__.py
+++ b/src/indra_cogex/apps/queries_web/__init__.py
@@ -83,29 +83,22 @@ CATEGORY_DESCRIPTIONS = {
 gene_expression_ns = Namespace("Gene Expression Queries", CATEGORY_DESCRIPTIONS['gene_expression'], path="/api")
 go_terms_ns = Namespace("GO Terms Queries", CATEGORY_DESCRIPTIONS['go_terms'], path="/api")
 clinical_trials_ns = Namespace("Clinical Trials Queries", CATEGORY_DESCRIPTIONS['clinical_trials'], path="/api")
-biological_pathways_ns = Namespace("Biological Pathways Queries", CATEGORY_DESCRIPTIONS['biological_pathways'],
-                                   path="/api")
+biological_pathways_ns = Namespace("Biological Pathways Queries", CATEGORY_DESCRIPTIONS['biological_pathways'], path="/api")
 drug_side_effects_ns = Namespace("Drug Side Effects Queries", CATEGORY_DESCRIPTIONS['drug_side_effects'], path="/api")
 ontology_ns = Namespace("Ontology Queries", CATEGORY_DESCRIPTIONS['ontology'], path="/api")
-literature_metadata_ns = Namespace("Literature Metadata Queries", CATEGORY_DESCRIPTIONS['literature_metadata'],
-                                   path="/api")
+literature_metadata_ns = Namespace("Literature Metadata Queries", CATEGORY_DESCRIPTIONS['literature_metadata'], path="/api")
 statements_ns = Namespace("Statements Queries", CATEGORY_DESCRIPTIONS['statements'], path="/api")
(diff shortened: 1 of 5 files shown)


src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -83,22 +83,29 @@
...
 gene_domains_ns = Namespace("Gene Domain Queries", CATEGORY_DESCRIPTIONS['gene_domains'], path="/api")
-phenotype_variant_ns = Namespace("Phenotype-Variant Association Queries", CATEGORY_DESCRIPTIONS['phenotype_variant'], path="/api")
+phenotype_variant_ns = Namespace("Phenotype-Variant Association Queries", CATEGORY_DESCRIPTIONS['phenotype_variant'],
+                                 path="/api")
 drug_indications_ns = Namespace("Drug Indication Queries", CATEGORY_DESCRIPTIONS['drug_indications'], path="/api")
 gene_codependence_ns = Namespace("Gene Codependence Queries", CATEGORY_DESCRIPTIONS['gene_codependence'], path="/api")
 enzyme_activity_ns = Namespace("Enzyme Activity Queries", CATEGORY_DESCRIPTIONS['enzyme_activity'], path="/api")
-cell_line_properties_ns = Namespace("Cell Line Property Queries", CATEGORY_DESCRIPTIONS['cell_line_properties'], path="/api")
+cell_line_properties_ns = Namespace("Cell Line Property Queries", CATEGORY_DESCRIPTIONS['cell_line_properties'],
+                                    path="/api")
```
kkaris, line 108: This diff looks spurious: please reverse this change since it doesn't do any material change and only adds line breaks

src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -320,9 +327,34 @@ def get_example_data():
...
+
 examples_dict = {
     "tissue": fields.List(fields.String, example=["UBERON", "UBERON:0001162"]),
-    "gene": fields.List(fields.String, example=["HGNC", "9896"]),
+    "gene": fields.List(fields.String,
+                        example=MultiExample(
+                            ["HGNC", "9896"],
+                            ["hgnc", "57"],
+                            ["hgnc", "10007"]
+                        ).get_example()),
```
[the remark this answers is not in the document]
prasham21: What if multiple functions are using the same example. 
For eg: the HGNC 9896 is utilized as an example by various functions.  

Should I assign a default value for most functions that use that parameter and only for the ones that are returning empty values, I should the function name to look up for their specific example?

src/indra_cogex/apps/queries_web/__init__.py
prasham21: Fixed in 44c2d2c3

src/indra_cogex/apps/queries_web/__init__.py
kkaris, line 1: The examples used in the API-docs are set in this code: https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/apps/queries_web/__init__.py#L491-L500 so that's where you need to change the code to update the examples.

src/indra_cogex/apps/queries_web/__init__.py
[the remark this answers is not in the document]
kkaris: Either use the default value for the example that most functions use, or just repeat the value with multiple entries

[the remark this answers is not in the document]
prasham21: Fixed in c58f2365

src/indra_cogex/apps/queries_web/__init__.py
prasham21: Fixed in c58f2365

kkaris:
Looks good to me

kkaris (approved):

PR #195, opened 2025-01-16
Add ability to browse statements underlying gene set analysis results

prasham21:
This PR adds the ability to view underlying INDRA statements for each upstream and downstream ORA results.

### DIFF
--- a/src/indra_cogex/apps/queries_web/__init__.py
+++ b/src/indra_cogex/apps/queries_web/__init__.py
@@ -406,6 +406,7 @@ examples_dict = {
     "negative_genes": fields.List(fields.String, example=EXAMPLE_NEGATIVE_HGNC_IDS),
     "gene_names": fields.List(fields.String, example=continuous_analysis_example_names),
     "target_id": fields.String(example="go:0006955"),
+    "is_downstream": fields.Boolean(example=False),
     "log_fold_change": fields.List(fields.Float, example=continuous_analysis_example_data),
     "species": fields.String(example="human"),
     "permutations": fields.Integer(example=100),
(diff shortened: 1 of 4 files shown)


src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -466,7 +472,8 @@ def get_example_data():
     [(subnetwork, fn) for fn in ["indra_subnetwork_relations", "indra_subnetwork_meta", "indra_mediated_subnetwork",
                                  "indra_subnetwork_tissue", "indra_subnetwork_go"]] +
     [(metabolite_analysis, fn) for fn in ["metabolite_discrete_analysis"]] +
-    [(gene_analysis, fn) for fn in ["discrete_analysis", "signed_analysis", "continuous_analysis"]]
+    [(gene_analysis, fn) for fn in ["discrete_analysis", "signed_analysis", "continuous_analysis"]] +
+    [(search, fn) for fn in ["get_ora_statements"]]
```
kkaris, line 476: When I test `get_ora_statements` in the rest API docs, it errors. Can you look into the examples there? We might have to add a special case for `get_ora_statements`

src/indra_cogex/apps/search/search.py
```diff
@@ -93,7 +102,7 @@ def search():
             paper_term=paper_id,
             mesh_term=mesh_terms,
             limit=1000,
-            evidence_limit=1000,
+            minimum_evidence=1000,
```
kkaris, line 105: Why is this changed? I think `evidence_limit` is the correct parameter name: https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/client/queries.py#L1236

prasham21: Fixed in 4848c023

kkaris (approved):

PR #198, opened 2025-02-06
Adding New Feature: Exposing Background genes

prasham21:
This brach deals with adding a new feature which allows users to enter a list of background genes(optional)  in the discrete GSEA. 
After obtaining the results, the user has the option to toggle and view all the background genes they had entered in the form field.


kkaris:
The example background genes are exactly the same as the gene list, and when I run the analysis with both the examples, there is no result. Try to find a list of example background genes for which there is a result.

bgyori (approved):

PR #202, opened 2025-02-20
Add Separate Kinase and TF Views in INDRA Upstream Analysis Results

prasham21:
Adds new functionality to view Kinase and Transcription Factor (TF) specific statements from INDRA Upstream analysis results. 
Kinases now display only phosphorylation statements, while TFs show increase/decrease amount statements. 
This enhancement improves the ability to analyze regulator-specific mechanisms in the upstream analysis.

### DIFF
--- a/src/indra_cogex/apps/templates/gene_analysis/discrete_results.html
+++ b/src/indra_cogex/apps/templates/gene_analysis/discrete_results.html
@@ -80,6 +80,8 @@
 
 {% macro render_indra_table(df, table_id) -%}
    {% set is_downstream = "downstream" in table_id %}
+   {% set is_kinase = "kinase" in table_id %}
+   {% set is_tf = "tf" in table_id %}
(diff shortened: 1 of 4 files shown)


src/indra_cogex/apps/templates/gene_analysis/discrete_results.html
```diff
@@ -20,11 +20,6 @@
             "order": [[2, "asc"]],
             pageLength: 10,
             layout: {
-                // Full documentation of layout:
```
kkaris, line 23: I think we can keep the comments for the DataTables settings, then you don't have to wander around on their homepage to find the right part in their documentation.

prasham21: Fixed in 87b08460
Also reinstated some previous minor changes

bgyori (approved):

PR #205, opened 2025-03-06
Search UI Enhancement

haohangyan:
This PR fixes the following things:
Make sure the evidence returned with the statement is filtered by mesh_term [done]
Show CURIE id in grounded result [done]
Support recognizing CURIE for agent input [done]
Change the statement browser page text [done]

### DIFF
--- a/src/indra_cogex/apps/search/search.py
+++ b/src/indra_cogex/apps/search/search.py
@@ -73,7 +73,7 @@ def search():
     # GET Request: Extract query parameters and fetch statements
     agent = request.args.get("agent")
 
-    #Check if the agnet is in CURIE form
+    #Check if the agent is in CURIE form
     if agent:
         agent = check_and_convert(agent)
 
@@ -84,7 +84,7 @@ def search():
 
     other_agent = request.args.get("other_agent")
 
-    # Check if the other_agnet is in CURIE form
+    # Check if the other_agent is in CURIE form
     if other_agent:
         other_agent = check_and_convert(other_agent)
 
--- a/src/indra_cogex/apps/search/search.py
+++ b/src/indra_cogex/apps/search/search.py
@@ -66,6 +66,7 @@ def search():
             "mesh_tuple": request.form.get("mesh_tuple"),
         }
         query_params = {k: v for k, v in query_params.items() if v}
+        print(form.mesh_terms.data, request.form.get('mesh_tuple'))
         return redirect(url_for("search.search", **query_params))
 
     # GET Request: Extract query parameters and fetch statements
@@ -92,7 +93,7 @@ def search():
     if mesh_tuple:
         source_db, source_id = json.loads(mesh_tuple)
         mesh_terms = (source_db, source_id)
-
+    print(mesh_terms, mesh_tuple)
     # Fetch and display statements
     if agent or other_agent or rel_types:
         statements, evidence_count = get_statements(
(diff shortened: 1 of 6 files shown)


src/indra_cogex/apps/search/search.py
```diff
@@ -66,16 +67,27 @@ def search():
             "mesh_tuple": request.form.get("mesh_tuple"),
         }
         query_params = {k: v for k, v in query_params.items() if v}
+        print(form.mesh_terms.data, request.form.get('mesh_tuple'))
         return redirect(url_for("search.search", **query_params))
 
     # GET Request: Extract query parameters and fetch statements
     agent = request.args.get("agent")
+
+    #Check if the agnet is in CURIE form
```
bgyori, line 76:
```suggestion
    #Check if the agent is in CURIE form
```

src/indra_cogex/apps/search/search.py
```diff
@@ -66,16 +67,27 @@ def search():
             "mesh_tuple": request.form.get("mesh_tuple"),
         }
         query_params = {k: v for k, v in query_params.items() if v}
+        print(form.mesh_terms.data, request.form.get('mesh_tuple'))
```
kkaris, line 70: Do we need the prints for any other purpose than debugging? If not then I think they should be deleted to not bloat the logs when running the service.

src/indra_cogex/client/queries.py
```diff
@@ -928,6 +928,7 @@ def get_evidences_for_stmt_hashes(
     client: Neo4jClient,
     limit: Optional[str] = None,
     remove_medscan: bool = True,
+    mesh_terms: Optional[List[str]] = None,
```
kkaris, line 931: Should a similar filter be set up for `get_evidences_for_stmt_hash` as well? I'm not sure if that's really needed, but something to discuss before merging.

kkaris:
Something else I found that isn't really connected to the scope of this PR but in the same code and is probably pretty easy to fix:

If searching something with a large amount of evidence, say [ACE2 as an agent and COVID19 as a mesh term](https://discovery.indra.bio/search/?agent=ACE2&agent_tuple=%5B%22HGNC%22,%2213557%22%5D&agent_role=object&other_role=subject&mesh_tuple=%5B%22MESH%22,%22D000086382%22%5D), the top statement (hash 11069144779908191) has a lot of evidence but it is exactly 1000 according to the badges. However, looking up this statement, the evidence count and source counts are 4345 and `{"sparser": 2819, "reach": 1516, "trips": 8, "signor": 2}`, respectively.

I think the problem comes from here: https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/client/queries.py#L1386
The evidence count should be whatever the relationship data says, not capped out at `evidence_limit`, which is just a parameter that determines the number of evidences to get.

Instead, I think doing it like it's done in [`subnetwork_go` with source counts](https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/client/subnetwork.py#L414-L416) is better: return the source counts instead of evidence counts, then the source counts can be turned into evidence counts in the `search` endpoint function in `search.py`. Then both `evidence_count` and `source_counts` can be passed into [`render_statements`](https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/apps/utils.py#L78-L88) to give the correct badges on the rendered page.

PR #213, opened 2025-03-27
Functionalities to check agent existence

haohangyan:
This PR adds the functionality to check the existence of an agent to see if it is in the cogex database.
Only the CURIE check is implemented for the search page, but we can still use check_agent_existence() to check name strings in potential future use cases.

### DIFF
--- a/src/indra_cogex/client/neo4j_client.py
+++ b/src/indra_cogex/client/neo4j_client.py
@@ -1029,6 +1029,30 @@ class Neo4jClient:
         )
         self.create_tx(create_query)
 
+    def check_curie_exists(self, agent:  Union[str, Tuple[str, str]]) -> bool:
+        """Check if an agent with the given id exists in the database.
+
+        Parameters
+        ----------
+        agent : Union[str, Tuple[str, str]]
+        The agent to check. Can be a string or a CURIE tuple.
+        Returns
+        -------
+        bool
+            True if the agent exists, False otherwise.
+        """
+        if isinstance(agent, tuple):
+            agent = norm_id(*agent)
+            clause = f"(n:BioEntity {{id: $agent}})"
+        else:
+            clause = f"(n:BioEntity {{name: $agent}})"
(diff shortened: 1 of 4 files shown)


src/indra_cogex/client/neo4j_client.py
```diff
@@ -77,6 +77,21 @@ def __del__(self):
         if self.driver is not None:
             self.driver.close()
 
+    def load_agent_cache(self):
+        """Load all agents into a dictionary cache for quick lookup."""
+        query = "MATCH (n:BioEntity) RETURN n.id AS id, n.name AS name"
```
bgyori, line 82: We should add a constraint that the BioEntity is connected (arbitrary direction) to an :indra_rel.

src/indra_cogex/apps/wsgi.py
```diff
@@ -46,6 +47,10 @@
 app.extensions[INDRA_COGEX_EXTENSION] = Neo4jClient()
 app.extensions[STATEMENT_CURATION_CACHE] = CurationCache()
 
+neo4j_client = Neo4jClient()
+agent_cache = neo4j_client.load_agent_cache()
```
bgyori, line 51: Instead of running this at each startup, we could have a file-based cache and add its creation to the Dockerfile at build time.

src/indra_cogex/apps/constants.py
```diff
@@ -57,7 +57,7 @@
 INDRA_COGEX_EXTENSION = "indra_cogex_client"
 STATEMENT_CURATION_CACHE = "curation_cache"
 SOURCE_BADGES_CSS = STATIC_DIR / "source_badges.css"
-
+AGENT_NAME_CACHE = APPS_DIR / "search/agent_cache.pkl"
```
bgyori:
```suggestion
AGENT_NAME_CACHE = APPS_DIR / "search" / "agent_cache.pkl"
```

PR #217, opened 2025-04-03
Clarify endpoints

kkaris:
This PR adds text to all analysis endpoint pages that spells out what their corresponding rest api endpoints are. It also adds mention of the form based pages for each of the backend analysis functions.


bgyori (approved):
Looks great!

PR #228, opened 2025-04-24
Fixes for Source-Target Gene Analysis: Enhance source_target_explanation.py to save individual artifacts and use Vue-compatible statements.

prasham21:
This PR addresses two main issues:

Updates the CLI for explain_downstream to write individual artifacts to the output folder instead of just a pickle file
Modifies statement handling to return structured data for Vue.js rendering instead of pre-generated HTML

These changes improve both the CLI experience and frontend integration by:

Making results directly accessible without loading the pickle file
Enabling interactive statement rendering with existing Vue components
Supporting REST API endpoints with structured statement data
Maintaining backward compatibility with the pickle file option

### DIFF
--- a/src/indra_cogex/analysis/source_targets_explanation.py
+++ b/src/indra_cogex/analysis/source_targets_explanation.py
@@ -717,7 +717,7 @@ def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, output_dir=
 
 
 @autoclient()
-def explain_downstream(
+def source_target_analysis(
     source: str,
     targets: List[str],
     output_dir: Optional[str] = None,
--- a/src/indra_cogex/analysis/source_targets_explanation.py
+++ b/src/indra_cogex/analysis/source_targets_explanation.py
@@ -14,10 +14,8 @@ import logging
 import itertools
 import os
 import pickle
-from collections import defaultdict
 from typing import List, Tuple, Optional, Dict, Any
 from io import BytesIO
-from bs4 import BeautifulSoup
 
 import pandas as pd
 import matplotlib
@@ -28,7 +26,6 @@ matplotlib.use('agg')
 import matplotlib.pyplot as plt
 
 from indra.databases import hgnc_client
-from indra.assemblers.html import HtmlAssembler
 from indra.statements import *
 
 from indra_cogex.client import *
@@ -247,11 +244,11 @@ def assemble_protein_stmt_htmls(stmts_df):
                 try:
                     stmt_jsons.append(json.loads(sj))
                 except Exception as e:
-                    print(f"Error decoding stmt_json string: {sj[:100]} -> {e}")
+                    logger.error(f"Error decoding stmt_json string: {sj[:100]} -> {e}")
             elif isinstance(sj, dict):
                 stmt_jsons.append(sj)
             else:
-                print(f"Skipping unexpected stmt_json type: {type(sj)} -> {sj}")
+                logger.warning(f"Skipping unexpected stmt_json type: {type(sj)} -> {sj}")
 
         stmts = stmts_from_json(stmt_jsons)
 
@@ -588,6 +585,7 @@ def run_explain_downstream_analysis(
     target_hgnc_ids,
     output_dir=None,
     curations=None,
+    minimal_output=False,
     *,
     client
 ):
@@ -606,6 +604,10 @@ def run_explain_downstream_analysis(
         of the web app, this will be done automatically, if this function is called
         standalone, curations need to be provided manually. In order to not provide
         any curations outside the context of the web app, pass an empty list.
+    minimal_output : bool
+        If True, removes HTML styled text and other rendering-specific data from
+        statement formatting. Should be set to True for REST API calls and False
+        for web form workflows.
     client :
         The client instance
 
@@ -644,12 +646,12 @@ def run_explain_downstream_analysis(
                 try:
                     stmt_jsons.append(json.loads(sj))
                 except json.JSONDecodeError as e:
-                    print(f"Failed to decode JSON string: {sj[:100]}... -> {e}")
+                    logger.warning(f"Failed to decode JSON string: {sj[:100]}... -> {e}")
                     continue
             elif isinstance(sj, dict):
                 stmt_jsons.append(sj)
             else:
-                print(f"Unexpected type in stmt_json column: {type(sj)} -> {sj}")
+                logger.warning(f"Unexpected type in stmt_json column: {type(sj)} -> {sj}")
                 continue
 
         stmts = stmts_from_json(stmt_jsons)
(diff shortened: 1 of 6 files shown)


src/indra_cogex/analysis/source_targets_explanation.py
```diff
@@ -531,15 +549,17 @@ def convert_plot_to_base64(fig):
 
 
 @autoclient()
-def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, *, client):
-    """Run complete downstream analysis.
+def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, output_dir=None, *, client):
+    """Run complete downstream analysis and save results to files.
```
kkaris, line 553:
```suggestion
    """Run complete downstream analysis and optionally save results to files.
```

src/indra_cogex/analysis/source_targets_explanation.py
```diff
@@ -551,27 +571,99 @@ def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, *, client):
...
-    results['statements'] = assemble_protein_stmt_htmls(filtered_df)
+    # Save interaction plot if output directory is specified
+    if output_dir:
+        # Save as PNG image
+        interaction_fig.savefig(os.path.join(output_dir, 'interaction_plot.png'))
+        # Also save the base64 data for convenience
+        with open(os.path.join(output_dir, 'interaction_plot.txt'), 'w') as f:
+            f.write(results['interaction_plot'])
+
+    # Process statements - Using the new Vue-compatible approach
```
kkaris, line 593:
```suggestion
    # Process statements
```

src/indra_cogex/analysis/source_targets_explanation.py
```diff
@@ -590,23 +695,58 @@ def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, *, client):
...
+        pathways_df.to_html(os.path.join(output_dir, 'combined_pathways.html'))
 
     # 8. Create analysis plots
     if not shared_go_df.empty and not shared_entities.empty:
         # graph_boxplots now returns base64 string directly
         results['analysis_plot'] = graph_boxplots(shared_go_df, shared_entities)
+        if output_dir:
+            # Save as base64 text file
+            with open(os.path.join(output_dir, 'analysis_plot.txt'), 'w') as f:
+                f.write(results['analysis_plot'])
```
kkaris, line 721: Again, I don't think we need to save the plots as base64 if we also are saving them as PNG.

src/indra_cogex/analysis/source_targets_explanation.py
```diff
@@ -590,23 +695,58 @@ def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, *, client):
...
+                from io import BytesIO
+                from PIL import Image
+
+                # Assuming the base64 string is for an image
+                img_data = base64.b64decode(results['analysis_plot'].split(',')[1])
+                img = Image.open(BytesIO(img_data))
+                img.save(os.path.join(output_dir, 'analysis_plot.png'))
+            except Exception:
+                # If conversion fails, just skip the PNG output
+                pass
```
kkaris, line 735: I think the error should be raised if it doesn't work, in any normal case, a plot should be able to be produced.

src/indra_cogex/analysis/source_targets_explanation.py
prasham21: Fixed in a93d455e

src/indra_cogex/analysis/source_targets_explanation.py
prasham21: Fixed in a93d455e

src/indra_cogex/analysis/source_targets_explanation.py
prasham21: Fixed in a93d455e

src/indra_cogex/analysis/source_targets_explanation.py
```diff
@@ -551,27 +571,99 @@ def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, *, client):
...
     results['interaction_plot'] = convert_plot_to_base64(interaction_fig)
 
-    results['statements'] = assemble_protein_stmt_htmls(filtered_df)
+    # Save interaction plot if output directory is specified
+    if output_dir:
+        # Save as PNG image
+        interaction_fig.savefig(os.path.join(output_dir, 'interaction_plot.png'))
+        # Also save the base64 data for convenience
+        with open(os.path.join(output_dir, 'interaction_plot.txt'), 'w') as f:
+            f.write(results['interaction_plot'])
```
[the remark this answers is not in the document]
prasham21: Fixed in a93d455e

src/indra_cogex/analysis/source_targets_explanation.py
prasham21: Fixed in a93d455e

kkaris:
Found a bug when running the rest api endpoint for source-target analysis. Looking into it.

prasham21:
> Found a bug when running the rest api endpoint for source-target analysis. Looking into it.

Is it one of these two errors `TypeError: unhashable type: 'list' ` or  `IndexError: index 0 is out of bounds for axis 0 with size 0.`
I remember resolving this error some time back.

kkaris:
I think the source-target web form results with the Vue components work well now and I only have two things that we might want to change:

1. There is a lot of whitespace between each row of statements (see screenshot), I think this can be fixed by simply removing the custom style rule for containers you have [here](https://github.com/prasham21/indra_cogex/blob/sourcetarget-page-update/src/indra_cogex/apps/templates/source_target/source_target_results.html#L23). 
image
2. The statement data that comes out is mostly fine, but we could tweak the data to skip some of the stuff that is purely for rendering purposes, like the HTML styled text. This could be done with special handling [here](https://github.com/prasham21/indra_cogex/blob/sourcetarget-page-update/src/indra_cogex/analysis/source_targets_explanation.py#L841-L847) by providing a flag to `source_target_analysis` where if it's true, fix the statement data structure being returned. Then provide that flag when `source_target_analysis` is called from the REST API but not when it's called when using the web form workflow.
image

src/indra_cogex/analysis/source_targets_explanation.py
```diff
@@ -551,27 +617,97 @@ def run_explain_downstream_analysis(source_hgnc_id, target_hgnc_ids, *, client):
...
+            if isinstance(sj, str):
+                try:
+                    stmt_jsons.append(json.loads(sj))
+                except json.JSONDecodeError as e:
+                    print(f"Failed to decode JSON string: {sj[:100]}... -> {e}")
+                    continue
+            elif isinstance(sj, dict):
+                stmt_jsons.append(sj)
+            else:
+                print(f"Unexpected type in stmt_json column: {type(sj)} -> {sj}")
```
kkaris, line 652: Maybe transform the `print` statements to `logger.warning` and/or `logger.error`
```suggestion
                    logger.warning(f"Failed to decode JSON string: {sj[:100]}... -> {e}")
                    continue
            elif isinstance(sj, dict):
                stmt_jsons.append(sj)
            else:
                logger.warning(f"Unexpected type in stmt_json column: {type(sj)} -> {sj}")
```

PR #233, opened 2025-05-22
Medscan fix for ora statements

kkaris:
This PR fixes a few cases where the removal of medscan evidence was hard coded or not propagated properly.

The issue would show up as statements with zero evidence and showing an internal server error when trying to get the evidence for those statements.


kkaris:
Need to fix the rest api before we can merge, hold on.

PR #243, opened 2025-10-08
 Fix viz-network initialization issues and Enable flexible subnetwork input parsing

prasham21:
This PR has two fixes: 

1) This PR resolves the intermittent issue where the network graph failed to render on the first page load but appeared correctly after refresh or in incognito mode.

2) This PR updates the Subnetwork Explorer form and route to allow more flexible user input. Users can now enter plain HGNC IDs (e.g., 1008), gene symbols (e.g., TP53), or CURIEs (e.g., hgnc:11998, chebi:15377) without triggering validation errors.
The strict CURIE check in NodesForm was relaxed, and input normalization is now handled by parse_node_list(), ensuring consistent backend behavior across all input formats.

### DIFF
--- a/src/indra_cogex/apps/curator/explorer_blueprint.py
+++ b/src/indra_cogex/apps/curator/explorer_blueprint.py
@@ -865,7 +865,6 @@ def subnetwork():
             title="Subnetwork Explorer",
             description=f"""\
             The subnetwork explorer shows statements between the following nodes.
-            {_database_text("Pathway Commons")}
             {EVIDENCE_TEXT}
             </p>
             <p>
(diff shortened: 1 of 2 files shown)


src/indra_cogex/apps/wsgi.py
```diff
@@ -4,9 +4,11 @@
 
 import logging
 import os
+from pathlib import Path
 
 from flask import Flask
 from flask_bootstrap import Bootstrap4
+from flask_session import Session
```
bgyori: We need to declare this as a new dependency

PR #258, opened 2026-01-16
Add query function and test to get pmids from hashes

haohangyan:
This PR adds the functionality to get pmids from the statement hash in the cogex query module. This new function is used in https://github.com/gyorilab/indra_db/pull/255

### DIFF
--- a/src/indra_cogex/client/queries.py
+++ b/src/indra_cogex/client/queries.py
@@ -821,7 +821,7 @@ def get_pmids_for_stmt_hash(stmt_hash: int, *, client: Neo4jClient):
     query = """
         MATCH (e:Evidence {stmt_hash: $hash})-[:has_citation]-(p:Publication)
         WHERE p.id STARTS WITH 'pubmed:'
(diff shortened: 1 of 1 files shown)


src/indra_cogex/client/queries.py
```diff
@@ -800,6 +801,33 @@ def get_pmids_for_mesh(
...
+        The statement hash to query.
+    client :
+        The Neo4j client.
+
+    Returns
+    -------
+    :
+        The PubMed IDs for the given MESH term
+    """
+    query = "MATCH (e:Evidence {stmt_hash: $hash}) RETURN e.evidence"
```
bgyori: Here we should actually just follow the `has_citation` links from the Evidence to a Publication node and get the PMID from there.
```suggestion
    query = "MATCH (e:Evidence {stmt_hash: $hash})-[:has_publication]-(p:Publication) RETURN p.id"
```
and then we have to process p.id-s like pubmed:11170639 to get a clean set of PMIDs.

PR #263, opened 2026-02-18
Docker update

kkaris:
This PR updates the Dockerfile to use Ubuntu 24 to resolve some issues related to new releases of upstream dependencies. The updates were informed by the current state of the [INDRA Dockerfile](https://github.com/gyorilab/indra/blob/master/docker/Dockerfile).

Other update:
- Set up a new fontawesome source package as the previous one seemed to work intermittently.

A new image has been built from this branch and is what is currently running on [discovery.indra.bio](https://discovery.indra.bio).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:24.04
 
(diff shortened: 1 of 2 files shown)


kkaris:
Replaced by #276.

PR #264, opened 2026-03-06
Added query for ev retrieval with embedding

koolgax99:
- Adds `get_evidences_for_embedding` query function to `client/queries.py` that retrieves top-k INDRA evidences by cosine similarity using a Neo4j vector index (`ev_embedding_index`)

- Returns a mapping from `stmt_hash` to a list of `(Evidence, cosine_similarity_score)` tuples, supporting semantic search over the evidence store with 384-dimensional embeddings (compatible with `all-MiniLM-L6-v2`)

- Registers the new function in the web API (`queries_web/__init__.py`) with appropriate field definitions for `query_embedding` (list of floats) and `k` (integer), including a realistic example embedding for the Swagger UI


src/indra_cogex/apps/queries_web/__init__.py
```diff
@@ -509,6 +510,66 @@ def get_example_data():
     "indication": fields.List(fields.String, example=["mesh", "D002318"]),
     # For EC
     "enzyme": fields.List(fields.String, example=["ec-code", "3.4.21.105"]),
+    # For embedding-based evidence retrieval (all-MiniLM-L6-v2, 384-dim)
+    # Example: "BRAF phosphorylates MAP2K1"
+    "query_embedding": fields.List(fields.Float, example=[
```
kkaris: The example embedding vector takes up a lot of real estate in the file. Is it possible to move this to a separate file in this module, like `constants.py` or a data file that is loaded when this module is imported?

koolgax99: Thanks I just moved it by adding a contstants.py in the same folder

PR #273, opened 2026-07-21
Incremental ingest via LOAD CSV

kkaris:
This PR adds a module for a node and relationship ingestion API around Cypher's `LOAD CSV` command. The module docstring has links to relevant documentation as well as example queries. Tests are added for the new functions.

New methods are added to the `Neo4jClient` to delete nodes and relationships. Node and relationships are matched by labels and types, respectively, to narrow down which nodes and relations are deleted.

### DIFF
--- a/src/indra_cogex/sources/partial_ingestion.py
+++ b/src/indra_cogex/sources/partial_ingestion.py
@@ -1,48 +1,83 @@
-"""An API for partial node ingestion via Cypher LOAD CSV.
+"""An API for partial node ingestion via Cypher's LOAD CSV.
 
-For node and edge files produced by the cogex processors with neo4j-admin import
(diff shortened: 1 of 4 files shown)


src/indra_cogex/sources/partial_ingestion.py
```diff
@@ -0,0 +1,686 @@
...
+    run_query :
+        If True, the query will be executed. If False, the query will be
+        returned as a string. Default: True.
+
+    Returns
+    -------
+    :
+        If ``run_query`` is False, returns the Cypher query string to ingest the
+        relationships. If ``run_query`` is True, runs the query and returns
+        None. Default: True.
```
kkaris, line 641:
```suggestion
        None.
```

PR #274, opened 2026-08-03
Landing page update

kkaris:
This PR updates the node and edge summaries on the landing page for discovery.indra.bio.

- Node and edge information is moved to dedicated JSON files, living in `indra_cogex/apps/static/`
- `indra_cogex/apps/templates/home.html` is updated:
  - There are now only two counters for total nodes and edges visible when page is loaded
  - A collapsible area holds detailed node and edge counters
  - Counter items are now slightly shrunk and placed in a grid with tighter margins to take up less space
  - Each item has a tooltip that adds extra details
  - Some reorganization:
    - No hard coding of node labels (because it's moved to the JSON)
    - Update JS to trigger the counter effect on expand

### DIFF
--- a/src/indra_cogex/apps/constants.py
+++ b/src/indra_cogex/apps/constants.py
@@ -46,6 +46,7 @@ edge_labels = {
     "variant_phenotype_association": "Variant Phenotype Associations",
     "has_activity": "Enzyme Annotations",
     "published_by": "Journal-Publisher Associations",
+    "codependent_with": "Gene knockout correlations",
 }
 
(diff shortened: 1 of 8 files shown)


kkaris:
I'm looking into the test failure.

_Update:_ the failure is due to the release of nltk 3.10.1. Maintainers are aware of it and seem to be on top of making a solution. Temporary fix would be to 'anti-pin' nltk with `nltk!=3.10.1`, otherwise simply wait for the nltk maintainers to make a new release.

